PR #1, opened 2019-05-31
E9 and 8x8 Font Support

carverh:
drivers/e9.c - E9 Hack
void vga_load_8x8_font() - Load 8x8 Font using BIOS INT 10h Function 1112h

### DIFF
--- a/lib/print.c
+++ b/lib/print.c
@@ -3,6 +3,7 @@
 #include <stdarg.h>
 #include <lib/print.h>
 #include <drivers/vga_textmode.h>
+#include <drivers/e9.h>
 
 static const char *base_digits = "0123456789abcdef";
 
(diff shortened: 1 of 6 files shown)


lib/print.c
```diff
@@ -3,6 +3,7 @@
 #include <stdarg.h>
 #include <lib/print.h>
 #include <drivers/vga_textmode.h>
+#include <drivers/e9.h>
```
iiSaLMaN:
```suggestion
```
Unnecessary.

avdgrinten:
The commit message could be more descriptive.

too-r:
Functions need to take a `void` parameter e.g ```int e9_read(void);```. This is important for correctness.

mintsuki:
I won't pull this because

a. The 8x8 font is unreadable, and you mixed it with potentially useful code (e9) just for the meme.
b. The 8x8 font will break the VGA textmode driver, you didn't even test it properly.

PR #2, opened 2019-06-01
Add simple disk library

iiSaLMaN:
Add a simple disk library: `read_sector` `write_sector` using bios interrupt calls.

### DIFF
--- a/lib/disk.c
+++ b/lib/disk.c
@@ -1,0 +1,48 @@
+#include <stdint.h>
+#include <stddef.h>
+#include <lib/disk.h>
+#include <lib/real.h>
+
+struct dap
+{
+    unsigned short size; // 16
+    unsigned short count;
+    unsigned short offset;
+    unsigned short segment;
+    unsigned long long lba;
+};
+
+struct dap dap = { 16, 0, 0, 0, 0 };
+struct rm_regs reg = { 0 };
+
+inline void setup_dap(int lba, int count, int off, int seg)
+{
+    dap.lba = lba;
+    dap.count = count;
+    dap.offset = off;
+    dap.segment = seg;
+}
+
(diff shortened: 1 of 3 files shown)


mintsuki:
Use the same coding style as the rest of the project, it's not hard.

too-r:
This is still inconsistent with the coding style in the rest of the project. Braces do not go on a newline in functions or struct definitions.

mintsuki:
What toor said, and also do not pass structs as arguments, use struct pointers. Also do not return structs.
This is also part of our "coding style" (or coding conventions, rather).

mintsuki:
Also, something you should keep in mind. Instead of translating 32 bit pointers to segment:offset, we want to be able to load the sector _wherever_ we want in the 32 bit address space. Therefore what you should do is load the sector at a fixed location in a buffer:
```c
static uint8_t sector_buf[512];
```
And then you memcpy the sector to its final location.
Since we're at an address <64k you can hardcode the segment to 0 in the dap, and point the offset to `sector_buf`.

mintsuki:
One last note, this:
```c
static struct dap
{
    unsigned short size;
    unsigned short count;
    unsigned short offset;
    unsigned short segment;
    unsigned long long lba;
};

static struct dap dap = { DAP_SIZE, 0, 0, 0, 0 };
```
can be written as
```c
static struct {
    unsigned short size;
    unsigned short count;
    unsigned short offset;
    unsigned short segment;
    unsigned long long lba;
} dap = { DAP_SIZE, 0, 0, 0, 0 };
```
And it will look prettier :)

iiSaLMaN:
> This is still inconsistent with the coding style in the rest of the project. Braces do not go on a newline in functions or struct definitions.

Woah, i didn't notice that lol

PR #4, opened 2019-07-08
Made error messages more specific

sakuranakamoto:


sakuranakamoto:
Not sure how you feel about the second commit. I noticed that you only call read_sectors and that read_sectors already saves GPR's so there was no need to also save them in read_sector if it never get's called by itself.

PR #6, opened 2020-04-21
Added ext2 subdirectory support

404meemr:
All files must have full paths, so `/qloader2.cfg` instead of `qloader2.cfg`


404meemr:
Files no longer need to have full paths.

PR #7, opened 2020-04-29
Add basic support for FAT32. 

Daouki:
Only short names and reading from the main directory are supported.


mintsuki:
Implemented by pull request #8

PR #8, opened 2020-05-01
Full FAT32 support

KrekBuk:
Full FAT32 filesystem support. Supports LFNs, subdirectories and fragmented files.


mintsuki:
Looks good to me, thank you very much for the contribution!

PR #9, opened 2020-05-02
FAT32 improvements.

KrekBuk:
In the old code read_partition was called once for every cluster, now when the clusters are not fragmented they will be read in one read_partition call which should significantly improve read speeds when reading larger non-fragmented files (16 times faster in the best case scenario).

I also fixed a bug I didn't notice yesterday when some files could be omitted by fat32_open_in and incorrectly reported as non-existent if the cluster size was not 1.


mintsuki:
Thanks a lot!

PR #10, opened 2020-05-27
Complete example of how qloader2 can be used

AtieP:
Hello, I made a folder with a complete example of how qloader2 can be used on a fresh baked kernel so new people won't strug (like me at the begging).


mintsuki:
I think this complete example perhaps belongs in another repo of its own or we could just add a link in the readme to some good tutorials that use qloader2.

PR #12, opened 2020-05-29
Add KASLR for PIE ELF kernels when using stivale

N00byEdge:
For now, only the `RELA` segments with relocations of type `R_X86_64_RELATIVE` are supported. This is the most common relocation though, for example, my kernel is over 1000 relocations and only use this one type of relocation. If another relocation type is desired, it should be easy to add.

Loading with a slide of e.g. 650MB causes the kernel to load at that offset in physical memory too. 
I'll leave this as something to be fixed when there at least is a simple vmm, as it would be a little bit of a pain to do the currently desired memory layout in the existing page tables.


mintsuki:
Thanks for the great PR!

PR #15, opened 2020-06-27
gcc -m32 doesn't work

carverh:


ethan4984:
just compiled and ran qloader2 with gcc -m32 and it worked

mintsuki:
It works.

PR #17, opened 2020-08-25
Mark the partition as bootable

atomicbitty:
Some BIOSes (eg. my laptop's) require at least one partition to be marked as bootable, this patch can save one from frustration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ and install qloader2, one can do:
 dd if=/dev/zero bs=1M count=0 seek=64 of=test.img
 parted -s test.img mklabel msdos
 parted -s test.img mkpart primary 1 100%
-parted -s loader.img set 1 boot on
+parted -s test.img set 1 boot on
 
 echfs-utils -m -p0 test.img quick-format 32768
 echfs-utils -m -p0 test.img import path/to/qloader2.cfg qloader2.cfg
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ and install qloader2, one can do:
 dd if=/dev/zero bs=1M count=0 seek=64 of=test.img
 parted -s test.img mklabel msdos
 parted -s test.img mkpart primary 1 100%
-parted -s test.img set 1 boot on
+parted -s test.img set 1 boot on # Workaround for buggy BIOSes
 
 echfs-utils -m -p0 test.img quick-format 32768
 echfs-utils -m -p0 test.img import path/to/qloader2.cfg qloader2.cfg
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ and install qloader2, one can do:
 dd if=/dev/zero bs=1M count=0 seek=64 of=test.img
 parted -s test.img mklabel msdos
 parted -s test.img mkpart primary 1 100%
+parted -s loader.img set 1 boot on
+
 echfs-utils -m -p0 test.img quick-format 32768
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,6 +53,8 @@ and install qloader2, one can do:
 dd if=/dev/zero bs=1M count=0 seek=64 of=test.img
 parted -s test.img mklabel msdos
 parted -s test.img mkpart primary 1 100%
+parted -s loader.img set 1 boot on
```
VisualDevelopment, line 56: The image is called test.img, not loader.img

VisualDevelopment (changes requested):

atomicbitty:
Sorry for the mistake

VisualDevelopment (approved):
:+1: Looks good to me!

mintsuki:
Thank you @atomicbitty and @VisualDevelopment and WOW I cannot believe how terrible BIOSes can be. BIOSes should not care about the partition table, but alas they do, and it can be very annoying.

I suggest it be specified with a comment that setting an active partition is a workarond for potentially buggy BIOSes rather than something that qloader2 actually cares about.

Other than that, LGTM

README.md
```diff
@@ -53,6 +53,8 @@ and install qloader2, one can do:
 dd if=/dev/zero bs=1M count=0 seek=64 of=test.img
 parted -s test.img mklabel msdos
 parted -s test.img mkpart primary 1 100%
+parted -s test.img set 1 boot on
```
VisualDevelopment, line 56:
```suggestion
parted -s test.img set 1 boot on # Workaround for buggy BIOSes
```

VisualDevelopment:
Also, @mintsuki wait till you see the BIOSes that need some weird jmp at the start to mark the boot sector as valid :^)

atomicbitty:
@mintsuki some bios are way worse, some set the boot disk number to 0x00 with hard disks (it actually happened with a laptop which I tested - laptops have horrible bioses) and it results to disk read error. I may send another pull request for that.

VisualDevelopment:
> @mintsuki some bios are way worse, some set the boot disk number to 0x00 with hard disks (it actually happened with a laptop which I tested - laptops have horrible bioses) and it results to disk read error. I may send another pull request for that.

I actually noticed a workaround for this in GRUB code

atomicbitty:
yeah, i've done this exact same workaround in my bootloader.

mintsuki:
> @mintsuki some bios are way worse, some set the boot disk number to 0x00 with hard disks (it actually happened with a laptop which I tested - laptops have horrible bioses) and it results to disk read error. I may send another pull request for that.

Sure feel free to send a PR for that too. That is incredibly weird though, AFAIK drive 0 is supposed to be the A: floppy disk drive. Could it be that the BIOS just loads up the bootsector thinking it's a floppy's?

atomicbitty:
> Could it be that the BIOS just loads up the bootsector thinking it's a floppy's?
Probably not, because I was explicitly saying to the bios to emulate the usb drive as HDD. The problem with most proprietary BIOSes is that they are only tested with Windows and maaaaaybe linux or BSD. And these OSes are already protected against dumb BIOSes, so the bugs are (almost) never noticed.

> BIOSes should not care about the partition table, but alas they do, and it can be very annoying.
All proprietary BIOSes I tested either required a partition table (HDDs) or a BPB (FDDs) in order to boot.

VisualDevelopment:
I hate hardware vendors lol

PR #18, opened 2020-08-26
Force IPV4 in toolchain script

atomicbitty:
I was gonna build a toolchain for qloader2 and wget simply hanged. This patch fixes it.

### DIFF
--- a/toolchain/make_toolchain.sh
+++ b/toolchain/make_toolchain.sh
@@ -28,7 +28,7 @@ mkdir -p build
 cd build
 
 if [ ! -f binutils-$BINUTILSVERSION.tar.gz ]; then
-    wget -4 https://ftp.gnu.org/gnu/binutils/binutils-$BINUTILSVERSION.tar.gz
+    wget -4 https://ftp.gnu.org/gnu/binutils/binutils-$BINUTILSVERSION.tar.gz # Force IPv4 otherwise wget hangs
 fi
 if [ ! -f gcc-$GCCVERSION.tar.gz ]; then
     wget -4 https://ftp.gnu.org/gnu/gcc/gcc-$GCCVERSION/gcc-$GCCVERSION.tar.gz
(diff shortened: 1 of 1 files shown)


toolchain/make_toolchain.sh
```diff
@@ -28,10 +28,10 @@ mkdir -p build
 cd build
 
 if [ ! -f binutils-$BINUTILSVERSION.tar.gz ]; then
-    wget https://ftp.gnu.org/gnu/binutils/binutils-$BINUTILSVERSION.tar.gz
+    wget -4 https://ftp.gnu.org/gnu/binutils/binutils-$BINUTILSVERSION.tar.gz
```
VisualDevelopment, line 31:
```suggestion
    wget -4 https://ftp.gnu.org/gnu/binutils/binutils-$BINUTILSVERSION.tar.gz # Force IPv4 otherwise wget hangs
```

toolchain/make_toolchain.sh
```diff
@@ -28,10 +28,10 @@ mkdir -p build
...
 fi
 if [ ! -f gcc-$GCCVERSION.tar.gz ]; then
-    wget https://ftp.gnu.org/gnu/gcc/gcc-$GCCVERSION/gcc-$GCCVERSION.tar.gz
+    wget -4 https://ftp.gnu.org/gnu/gcc/gcc-$GCCVERSION/gcc-$GCCVERSION.tar.gz
```
VisualDevelopment, line 34:
```suggestion
    wget -4 https://ftp.gnu.org/gnu/gcc/gcc-$GCCVERSION/gcc-$GCCVERSION.tar.gz # Same as above
```

VisualDevelopment (commented):
Maybe add a comment why this was done for documentation purposes

VisualDevelopment (approved):
LGTM. :heavy_check_mark:

mintsuki:
May I have some rationale as to why wget would hang with IPv6 on this specific URL? Is it a GNU mirror problem?

atomicbitty:
It's probably that my something related to my Wi-Fi card or something related to my ISP.

mintsuki:
I will merge this for the time being as it does not hurt, but we should probably find a better solution to the problem, eventually.

PR #19, opened 2020-08-26
Better firmware compatibility.

atomicbitty:
I added some workarounds for buggy BIOSes and made the bootsector messages smaller because it lacked space.


src/bootsect/bootsect.asm
```diff
@@ -14,7 +17,25 @@ start:
...
+    test dl, 0x80
+    jz .fix_drive
+
+    ; Drive numbers from 0x80..0x8f should be valid
+    test dl, 0x70
+    jz .continue
+
+  .fix_drive:
+    ; Try to fix up the mess the BIOS have done
+    mov dl, 0x80
```
mintsuki, line 36: Could be worth reading back up some flag to make sure the new drive is not a bogus one and gracefully handle that scenario?

mintsuki:
LGTM, will merge for now.

src/bootsect/bootsect.asm
atomicbitty: sorry, I didn't understand exactly what you mean. Are you saying about testing if the new drive is valid.

mintsuki: What I meant to say is whether we should check for a signature after loading the other sectors in, other than just checking for a successful call to int 13h, to make sure we are reading from the correct drive (0x80) and not some other unrelated drive/bogus data.

PR #26, opened 2020-09-18
32bit stivale was not pushing a return address.

sleepy-monax:
When trying to boot skift using limine I was still getting an invalid address for the stivale struct.
Pushing an additional 0 where the return address should have been fixed the issue.


mintsuki:
Ah, thanks for the bug fix!

PR #60, opened 2020-12-29
toolchain: build toolchain with multiples process instead of 1 process

Supercip971:
it is way faster to build the toolchain that way﻿


Mintsuki:
Hi, thanks for the PR.

I am going to go ahead and close this PR since this is already done by running `make toolchain` from the root of the repo: https://github.com/limine-bootloader/limine/blob/unstable/Makefile#L35

As for `make_toolchain.sh`, it takes the `MAKEFLAGS` as its first argument, so running `make_toolchain.sh -j$(nproc)` will have the desired effect (which is what the `Makefile` in the root does).

Issue #64, opened 2021-02-01
Add support for ISO9660


jlxip:
Add support for ISO9660 so it can boot from a live CD. I'll try to do it myself in the next few days if I have the time.

jlxip:
Implemented in #69 ( ͡° ͜ʖ ͡°)

PR #67, opened 2021-02-21
iso9660: first commit

jlxip:
Implements #64


stage2/lib/part.c
```diff
@@ -254,7 +254,7 @@ static size_t volume_index_i = 0;
 void volume_create_index(void) {
     size_t volume_count = 0;
 
-    for (uint8_t drive = 0x80; drive < 0x8f; drive++) {
+    for (uint8_t drive = 0x80; drive; drive++) {
```
jlxip, line 257: This change wasn't necessary in the end. Should be reverted.

PR #71, opened 2021-02-22
misc: Keep stages.bin for cdboot

jlxip:
Lowkey forgot about that :clown_face:, since `iso9660-test` requires the target `bootloader`


jlxip:
Not needed anymore

PR #95, opened 2021-07-26
Make sure the requested stack is aligned at word boundary

Andy-Python-Programmer:
This pull request adds a check that the requested stack is aligned at a word boundary. If the stack is not aligned it makes things slower and also if you want to access a variable on stack it must be aligned if you are using rust.

### DIFF
--- a/stage23/protos/stivale.c
+++ b/stage23/protos/stivale.c
@@ -174,6 +174,12 @@ void stivale_load(char *config, char *cmdline) {
         print("stivale: Requested stack at: %X\n", stivale_hdr.stack);
     }
 
+    // Check if the requested stack is correctly aligned at a word
+    // boundary (8 bytes on x64 and 4 bytes on x86).
+    if ((stivale_hdr.stack & (sizeof(size_t) - 1)) != 0) {
+        print("warn: Requested stack is not aligned at word boundary.");
+    }
+
     stivale_struct.module_count = 0;
     uint64_t *prev_mod_ptr = &stivale_struct.modules;
     for (int i = 0; ; i++) {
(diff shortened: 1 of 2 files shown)


stage23/protos/stivale.c
```diff
@@ -174,6 +174,12 @@ void stivale_load(char *config, char *cmdline) {
         print("stivale: Requested stack at: %X\n", stivale_hdr.stack);
     }
 
+    // Check if the requested stack is correctly aligned at a word
+    // boundary (8 bytes on x64 and 4 bytes on x86).
+    if (stivale_hdr.stack % sizeof(size_t) != 0) {
```
pitust: This modulo could be changed into an and (which is a lot faster).

stage23/protos/stivale2.c
```diff
@@ -191,6 +191,12 @@ void stivale2_load(char *config, char *cmdline, bool pxe, void *efi_system_table
         print("stivale2: Requested stack at: %X\n", stivale2_hdr.stack);
     }
 
+    // Check if the requested stack is correctly aligned at a word
+    // boundary (8 bytes on x64 and 4 bytes on x86).
+    if (stivale2_hdr.stack % sizeof(size_t) != 0) {
```
pitust: see above.

stage23/protos/stivale.c
Andy-Python-Programmer: Fixed!

stage23/protos/stivale2.c
Andy-Python-Programmer: Fixed!

pitust:
This message is gonna show for a total of like a tenth of a second, maybe there could be a feature where it would sleep for one second so that the user can actually acknowledge the issue.

PR #98, opened 2021-08-04
PE32/PE64 Support

AtieP:
For the Windows shills! :D

Sadly, no KASLR and relocations support, since as far as I am aware PE doesn't support it natively and requires some hacks. There's also a problem with the `.stvl` section. PE seems to align sections to 32, and the stivale header is only 24 bytes big.

### DIFF
--- a/stage23/lib/elf.c
+++ b/stage23/lib/elf.c
@@ -115,14 +115,47 @@ struct elf64_rela {
     uint64_t r_addend;
 };
 
-int elf_bits(uint8_t *elf) {
+static bool elf_validate(uint8_t *elf, bool is_64) {
     struct elf64_hdr hdr;
-    memcpy(&hdr, elf + (0), 20);
+    memcpy(&hdr, elf + (0), sizeof(struct elf64_hdr));
 
     if (strncmp((char *)hdr.ident, "\177ELF", 4)) {
         print("elf: Not a valid ELF file.\n");
-        return -1;
+        return false;
     }
+
+    if (hdr.ident[EI_DATA] != BITS_LE) {
+        print("elf: Not a Little-endian ELF file.\n");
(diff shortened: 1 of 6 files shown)


InsaneMiner:
I know Im not a contributor, but i think it should have a bit more work on it before it gets merged, that's only my opinion tho.

seriocomedy:
As a PE shill I say this is both COOL and GOOD.

InsaneMiner:
> As a PE shill I say this is both COOL and GOOD.

yes it is very cool

Neptune650:
> I know Im not a contributor, but i think it should have a bit more work on it before it gets merged, that's only my opinion tho.

What you mean by that?

InsaneMiner:
> > I know Im not a contributor, but i think it should have a bit more work on it before it gets merged, that's only my opinion tho.
> 
> What you mean by that?

i could be wrong, but i think it should at least support multiboot1. I know some people that like multiboot and use windows, so i think they would like that. (this is only my opinion)

Andy-Python-Programmer:
@InsaneMiner limine already has work in progress multiboot 1 support :^)

image

InsaneMiner:
> @InsaneMiner limine already has work in progress multiboot 1 support :^)

i know, but does the PE thing support it. I could just be completely wrong tho.

AtieP:
> > @InsaneMiner limine already has work in progress multiboot 1 support :^)
> 
> i know, but does the PE thing support it. I could just be completely wrong tho. 

PE support can't be added for multiboot because we can't modify the standard.

Andy-Python-Programmer:
You could make a RFC to multiboot.

AtieP:
> You could make a RFC to multiboot.

no thanks, that's called stivale(2)

PR #112, opened 2021-09-21
Override default higher-half offset address

ulmer-a:
Currently the default load offset address for higher-half kernels is hardcoded in `stage23/lib/elf.h`. This PR makes the hardcoded value a default and adds a configuration option to override it. 

I use it to let my own kernel load at an address near `0xffff800000000000`, but I thought maybe other people would like to configure this for themselves as well.


Mintsuki:
There is a reason behind the stivale specification dictating that kernels are to be loaded in the -2GiB area: it allows for use of `-mcmodel=kernel` which optimises kernels to be ran at this location (for non PIE kernels), and it allows a clean separation between the kernel's VMA area and the direct map.

Since there is no real reason to load the kernel image so low in the higher half (`0xffff800000000000`) I would suggest you to simply conform to the stivale spec (and do practically what all kernels do) and load your kernel in the -2GiB area.

ulmer-a:
Ok, anyway thanks for the info! I'll probably do that now.

PR #113, opened 2021-09-28
NTFS Support

Itay2805:
This PR adds basic NTFS support, it fully works with mkfs.ntfs but I have not tested it with ntfs created by windows.

It does not support compression nor encryption, and it does not attempt to use the fixup sequences.


Mintsuki:
Hi thanks for the PR!

From a quick skim it looks good to me. I will merge this for now, then feel free to fix possible bugs we discover down the line and PR those fixes.

PR #114, opened 2021-09-29
NTFS Improvements

Itay2805:
Started testing the driver on windows ntfs disk. Added support for resident directories and fixed bug with runlist parsing. We are still missing resident file data but that will hopefully come in another PR very soon :)


Mintsuki:
Thanks!

PR #115, opened 2021-10-04
multiboot2: add the EFI image handle tag

Andy-Python-Programmer:
This pull request adds the EFI image handle tag in multiboot2,


Mintsuki:
LGTM

PR #116, opened 2021-10-04
acpi_get_rsdp: UEFI RSDP lookup return RSDPv1 if avaliable

Andy-Python-Programmer:
This pull request makes the `acpi_get_rsdp` function to return the RSDPv1 if
avaliable (in UEFI) to match this functions behaviour under BIOS where
it returns the XSDP or the RSDP depending on which one is avaliable.


Mintsuki:
LGTM

PR #120, opened 2021-10-13
Issue with gnu-efi url

1dime:
Original URL didnt work. Also, run export GIT_SSL_NO_VERIFY=1 as it seems there's something wrong on sourceforge's end.


Mintsuki:
I frankly don't see how this fixes anything. Seems to work fine on my end, but I'm going to go ahead and import the gnu-efi repo into the Limine gh org to avoid issues with sourceforge in the future.

PR #123, opened 2021-11-05
stivale+stivale2: default to file path as the module string if NULL

Andy-Python-Programmer:
This pull request adds (stivale+stivale2):

* If a module string is not specified for a module then it defaults to use the file path as the module string instead. Instead of just setting it to `NULL`. Since just specifying an empty string is not a good indication but specifying the path is actually more saner.

* An alternate solution would be to pass the file name instead of the file path but I picked to use the full file path since eh its less conflict able with similar file modules.


ilobilo:
I think file name instead of path would be better
And if there are two files with same name then use full path

Andy-Python-Programmer:
@ilobilo That would not be consistent

PR #124, opened 2021-11-06
bootboot: add initrd support, and add a warning about the init stack

pitust:
This PR finishes bootboot support, for now at least, by adding support for the bootboot initrd things!

This means bootboot support is complete!

The initrd formats supported are:
 - cpio (oldc, newc and maybe crc)
 - ustar
 - raw kernel elf file that gets loaded directly

The KERNEL_PATH variable is treated as INITRD_PATH, except when INITRD_PATH is set too, when it overrides the kernel in the initrd.

Where does the kernel and initrd come from?
| KERNEL_PATH | INITRD_PATH | initrd        | kernel    |
|-------------|-------------|---------------|---------------|
| no          | no          | N/A           | N/A           |
| elf         | no          | kernel_path   | kernel_path   |
| elf         | any         | initrd_path   | kernel_path   |
| archive     | any         | N/A           | N/A           |
| no          | archive     | initrd_path   | initrd        |
| no          | elf         | initrd_path   | initrd_path   |

initrd_path: the initrd binary
initrd: the kernel fine inside of the initrd
kernel_path: the kernel binary
N/A: doesn't work

### DIFF
--- a/stage23/protos/bootboot.c
+++ b/stage23/protos/bootboot.c
@@ -186,7 +186,7 @@ void bootboot_load(char *config) {
 
     if (init_stack_size == ~0UL) {
         print("bootboot: warning: no init stack size entered, assuming 1024\n");
-        print("1024 is really small, specify more using ");
+        print("1024 is really small, specify more using initstack=size ini initrd;\n");
         init_stack_size = 1024;
     }
     printv("bootboot: mapping struct to %X\n", struct_vaddr);
--- a/stage23/protos/bootboot.c
+++ b/stage23/protos/bootboot.c
@@ -135,8 +135,6 @@ void bootboot_load(char *config) {
 
     /// Memory mappings ///
     pagemap_t pmap = new_pagemap(4);
-    BOOTBOOT* bootboot = (BOOTBOOT*)ext_mem_alloc_type_aligned(4096, MEMMAP_BOOTLOADER_RECLAIMABLE, 4096);
-    map_page(pmap, struct_vaddr, (uint64_t)(size_t)bootboot, VMM_FLAG_PRESENT | VMM_FLAG_WRITE, false);
 
     /// Load kernel ///
     uint64_t entry, top, slide, rangecount, physbase, virtbase = 0;
@@ -193,6 +191,11 @@ void bootboot_load(char *config) {
     printv("bootboot: mapping environemnt to %X\n", env_vaddr);
     printv("bootboot: mapping framebuffer to %X\n", fb_vaddr);
     printv("bootboot: the init stack is %X bytes\n", init_stack_size);;
+
+    /// Bootboot structure ///
+    BOOTBOOT* bootboot = (BOOTBOOT*)ext_mem_alloc_type_aligned(4096, MEMMAP_BOOTLOADER_RECLAIMABLE, 4096);
+    map_page(pmap, struct_vaddr, (uint64_t)(size_t)bootboot, VMM_FLAG_PRESENT | VMM_FLAG_WRITE, false);
+
 
     /// Environment ///
     {
--- a/stage23/protos/bootboot.c
+++ b/stage23/protos/bootboot.c
@@ -2,6 +2,7 @@
 #include <stddef.h>
 #include <stdbool.h>
 #include <protos/bootboot.h>
+#include <protos/bootboot/initrd.h>
 #include <lib/libc.h>
 #include <lib/elf.h>
 #include <lib/blib.h>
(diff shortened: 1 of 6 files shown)


leap0x7b:
when initrd is needed :nooo:

pitust:
> when initrd is needed :nooo:

you don't! set the elf to kernel_path and it should just work:tm:

leap0x7b:
why load elf as initrd

leap0x7b:
~~its not an archive format~~

pitust:
> why load elf as initrd

(a) that's what bootboot does, kinda
(b) this means you can have _just_ the kernel, no initrd

pitust:
Also, the kernel may want debug symbols or some shit like that from there. ELF can contain many sections with data, close enough.

bztsrc:
I feel some disturbance in the force... I think this compatibility matrix is heavily overcomplicated and totally unnecessary.

In the BOOTBOOT Potocol, there's only one file that gets booted. That file can be
- a single statically linked executable with all the blobs (like the Minix kernel for example)
- an initrd with multiple files, in which case the kernel must be one of the files in the archive

So it doesn't matter which variable you're using (KERNEL_PATH or INITRD_PATH) to specify that one file to boot, but you'll only need one of those variables for sure, and there's no need for this compatibility matrix at all.

Cheers,
bzt

pitust:
> I feel some disturbance in the force... I think this compatibility matrix is heavily overcomplicated and totally unnecessary.
> 
> In the BOOTBOOT Potocol, there's only one file that gets booted. That file can be
> 
> * a single statically linked executable with all the blobs (like the Minix kernel for example)
> * an initrd with multiple files, in which case the kernel must be one of the files in the archive
> 
> So it doesn't matter which variable you're using (KERNEL_PATH or INITRD_PATH) to specify that one file to boot, but you'll only need one of those variables for sure, and there's no need for this compatibility matrix at all.
> 
> Cheers, bzt

Hmm, i may have overcomplicated my explanation. The behaviour is pretty simple: the initrd is KERNEL_PATH, unless INITRD_PATH is specified, and the kernel is loaded from the initrd, unless both are specified, when it's loaded from KERNEL_PATH.

This means you don't need to recreate archives every time.

Mintsuki:
Is supporting ustar and cpio *really* a necessity with BOOTBOOT? This pr seems a bit overcomplicated to me

pitust:
> Is supporting ustar and cpio _really_ a necessity with BOOTBOOT? This pr seems a bit overcomplicated to me

Yes, it is - it's core to the bootboot protocol. That's the only part of the implementation the previous pr didn't have (also all the bugs i fixed)

pitust:
> Is supporting ustar and cpio _really_ a necessity with BOOTBOOT? This pr seems a bit overcomplicated to me

heh, regular bootboot also has three others: FS/Z, SFS and EchFS

stage23/protos/bootboot.c
```diff
@@ -101,34 +182,22 @@ void bootboot_load(char *config) {
...
-        if (elf64_load(
-            kernel, &entry, &top, &slide, MEMMAP_KERNEL_AND_MODULES,
-            false, false, &ranges, &rangecount, true, &physbase, &virtbase)) {
-            panic("bootboot: elf64 load failed");
-        }
-        for (uint64_t mapvirt = virtbase, mapphys = physbase; mapphys < top;mapvirt += 0x1000, mapphys += 0x1000) {
-            map_page(pmap, mapvirt, mapphys, VMM_FLAG_PRESENT | VMM_FLAG_WRITE, false);
-        }
+    if (init_stack_size == ~0UL) {
+        print("bootboot: warning: no init stack size entered, assuming 1024\n");
```
Andy-Python-Programmer, line 186: Suggestion: Maybe instead of making the stack too small, make it a reasonable about (like `4096 * 16`) and just warn that:
```
bootboot: WARNING: No init stack size specified, defaulting to `0x10000` (consult CONFIG.md). 
````

Also a note that warning will just be visible for a short second of time, maybe add a short delay so the warning is actually visible?

Comment (pitust) [reply stage23/protos/bootboot.c:186]: The spec says 1024, so i put 1024. But i will add the delay

Comment (Andy-Python-Programmer) [reply stage23/protos/bootboot.c:186]: Alrighty

**Comment by bztsrc** [comment]
> >    Is supporting ustar and cpio really a necessity with BOOTBOOT? This pr seems a bit overcomplicated to me

> Yes, it is - it's core to the bootboot protocol.

No, it is absolutely not, as a matter of fact quite the contrary. I quote from the spec page 11:

> As mentioned before, the BOOTBOOT Protocol does not specify the initial ramdisk format

and

> The protocol expects that a BOOTBOOT compliant loader iterates on the list of drivers until one returns a valid result.

> If all the file system drivers failed and returned {NULL,0}, a fallback driver will be initiated. That fallback driver will scan the ramdisk for the first file which has a valid executable format for the architecture.

There is one important part of the specification that your patch fails to comply with: if the format of the initrd isn't recognized, a BOOTBOOT compatible loader must search for the first executable for the architecture in the initrd's buffer. The executable must be contiguous in this case (fragmented files only supported through file system drivers, but not with the fallback search mechanism).

> 

Cheers,
bzt

**Comment by bztsrc** [comment]
I wrote this to you [in your ticket](https://gitlab.com/bztsrc/bootboot/-/issues/56#note_722040276):

> If you don't want to implement filesystem drivers, then it is enough if you search for the first executable for the platform (without fs drivers, the kernel must be contiguous inside the initrd). The spec expects that a loader uses different fs drivers, and then fallbacks to this search method. So if you don't have any fs drivers, it is enough to implement that latter fallback method to be compatible with the spec.

Hope this is clear now.
bzt

**Comment by pitust** [comment]
> > > Is supporting ustar and cpio really a necessity with BOOTBOOT? This pr seems a bit overcomplicated to me
> 
> > Yes, it is - it's core to the bootboot protocol.
> 
> No, it is absolutely not, as a matter of fact quite the contrary. I quote from the spec page 11:

Well yeah but if i am already wrote a driver it doesn't matter too much now does it?

**Comment by bztsrc** [comment]
> Well yeah but if i am already wrote a driver it doesn't matter too much now does it?

Yes, it is most certrainly better to have fs drivers than not having. But [this part](https://github.com/limine-bootloader/limine/blob/trunk/stage23/protos/bootboot/initrd.c#L19):
```c
    if (!memcmp(file.data, "\177ELF", 4)) {
```
must be the last fallback option not the first, and it must look for the header in the entire ramdisk buffer, not just in the first position.

Something like this (note I've added "else"s and in the last branch a "for"):
```c
if (file.data[4] == 0xBF) {
...
} else
if (!memcmp("07070", file.data, 5)) {
...
} else
if (!memcmp("ustar", file.data + 257, 5)) {
...
} else {
    for (file.data = initrd_ptr; file.data < initrd_ptr + initrd_size - 4; file.data++) {
        if (!memcmp(file.data, "\177ELF", 4)) {
            ...
        }
    }
}
```
Cheers,
bzt

Mintsuki:
@bztsrc I hope commit 1be24c3 fixes the concerns brought up above

bztsrc:
> @bztsrc I hope commit 1be24c3 fixes the concerns brought up above

Sadly no, there are several problems with it. But it is a big step toward a good implementation.

First, it only searches for the kernel if that has a specific name. That's not good, the only criteria for the search is that the initrd's format isn't recognized, the kernel's file name doesn't matter in this case. Second, it only checks every 4th offsets, again, not good, it must check every address, because no one knows if in an unknown archive format the header length is multiple of 4 bytes or not.

(Let's assume for a moment that there are no fs drivers, just this fallback search: this implementation would work for tar, which has a fixed length, multiple of 4 bytes header, but most likely would fail for cpio, which has a variable length header and does not guarantee that file contents start at 4 byte aligned addresses.)

Cheers,
bzt

bztsrc:
I forgot to mention yet another issue: you have to check for the architecture too, because the initrd might contain multiple kernels, one for each supported platform.

Here's a correct implementation for initrd.c (only the end of the file listed, from line 38, the beginning is ok):
```c
...
    }

    for (size_t i = 0; i < file.size - 19; i++) {
        if (memcmp(file.data + i, "\177ELF", 4) == 0 &&
            file.data[i + 18] == 62 && file.data[i + 19] == 0 /* ehdr->e_machine == EM_X86_64 */) {
                printv("bootboot: using autodetected kernel in initrd\n");
                return (struct initrd_file){
                    .size = file.size - i,
                    .data = file.data + i
                };
        }
    }

    panic("bootboot: cannot read file `%s`: cannot detect initrd type (only ustar, cpio and jamesm is supported).", path);
}
```
Cheers,
bzt

Mintsuki:
@bztsrc Do the 2 commits above make sense to you?

bztsrc:
> Do the 2 commits above make sense to you?

Well, to be perfectly honest with you, not really.

If you ask if this complies with the spec then the answer is probably yes, however this patch looks like you're your own enemy (no offense, please take this as a friendly advice).

1. there's a potential buffer overflow issue in bruteforce_kernel(). See above, the "for" loop runs to `file.size - 19`.
2. there's no point in having a known_initrd_format() check and using two separate code paths. This not only complicates your code unneccessarily, but also makes adding new initrd formats more difficult (you check for all magic twice, both in the source code and in run-time).
 
My advice is: remove  known_initrd_format(), keep only the first code path in bootboot.c without the unnecessary code duplication
```diff
-   if (known_initrd_format(bootboot_initrd_file)) {
        const char *corefile = config_get_value((char *)env, 0, "kernel");
        if (corefile == NULL) {
            corefile = "sys/core";
        }
        struct initrd_file file = initrd_open_auto(bootboot_initrd_file, corefile);
        if (!file.size) {
            panic("bootboot: cannot find the kernel!");
        }
        kernel = file.data;
-   } else {
-       struct initrd_file file = bruteforce_kernel(bootboot_initrd_file);
-       if (!file.size) {
-           panic("bootboot: cannot find the kernel!");
-       }
-       kernel = file.data;
-   }
```
and just look for the kernel at the end of initrd_open_auto() in initrd.c
```diff
- return (struct initrd_file){0};
+ return bruteforce_kernel(file);
```
This will make the code simpler and much more manageable on the long run. (And maybe this is just me, but I would move that two functions in initrd.c into bootboot.c, leaving only the file system drivers in the bootboot/ directory. But this is totally just personal preference.)

About all the other modifications in bootboot.c: I guess ok. There's no way to specify the environment file now, which is not against the BOOTBOOT Protocol per se, it's just it cripples the usability of limine considerably. But this is entirely up to the loader how to implement, where to load the environment from. You could copy a part of limine.cfg for the selected option for example. Loading it from a file inside the initrd most certainly a possible and working solution.

Cheers,
bzt

Mintsuki:
The environment is being loaded from sys/config. Is that not the 'preferred' way?

bztsrc:
Adding a new format still needs modifications at multiple different places, but that's your decision. At least this commit now complies with the spec, so no probs there!

> The environment is being loaded from sys/config. Is that not the 'preferred' way?

Nope, that is just a fallback option in the reference implementations. By default, they try to load FS0:\BOOTBOOT\CONFIG. As I've said, it is entirely up to the loader where to get the environment from, only the part how and what to pass to the kernel is mandated. You are free to implement whatever storage method you want. 

For example, there could be one KERNEL_ENVFILE per boot option in limine.cfg, or the already existing KERNEL_CMDLINE could be converted into a string with newline separated "key=value" pairs, entirely up to you how to do it. Looking for a file in the initrd is perfectly fine too.

Cheers,
bzt

PR #132, opened 2021-11-11
fix limine-install on BIOS disks

alicavus:
Without `limine-hdd.bin` limine-install is unsuccessfull:
```
$ sudo limine-install /dev/sda
Physical block size of 512 bytes.
ERROR: Could not determine if the device has a valid partition table.
       Please ensure the device has a valid MBR or GPT.
       Alternatively, pass `--force-mbr` at the end of the command to
       override these checks. ONLY DO THIS AT YOUR OWN RISK, DATA LOSS
       MAY OCCUR!
$ sudo limine-install --force-mbr  /dev/sda
ERROR: No such file or directory
```


Mintsuki:
@alicavus the reason you are getting the "No such file or directory" error is that you need to pass --force-mbr as *the last* argument to `limine-install`, as clearly stated in the first error message.

Issue #140, opened 2021-12-01
Circular dependency between PMM, VGA driver, and Panic.


lukflug:
**Limine version used**
Commit `92d4246` at [lukflug](https://github.com/lukflug/limine):trunk.
Only difference from Commit `0228c06` is that I've added `legacy-fat16-test` and `mbrhdd.bin` directives to the `Makefile`, in order to test it.

**Expected behavior**
`panic("Memory allocation failure.")`

**Actual (broken) behaviour**
Recursively calls `panic`. Evident as bootloop, presumably due to triple faulting due to a stack overflow or something like that.

**To Reproduce**
x86Box:
* Slot 1
* [i440FX] ASUS P/I-P65UP5 (C-PKND)
* Intel Pentium II (Klamath) @ 100MHz
* 16MB RAM
* [PCI] ATI Mach64GX with 4MB VRAM
* [PCI] IDE Controller (Dual-Channel)
* Hard Drive: `mbrtest.hdd` with CHS=130/16/63

**Stacktrace**
As there is no panic, I can't provide a stacktrace, but I was able to manually track down the program flow:
```
entry (entry.s2.c: 98)
init_memmap (pmm.s2.c: 269)
memmap_alloc_range (pmm.s2.c: 787)
panic (panic.s2.c: 22)
term_textmode (term.s2.c: 164)
init_vga_textmode (vga_textmode.s2.c: 143)
ext_mem_alloc (pmm.s2.c: 588)
ext_mem_alloc_type (pmm.s2.c: 592)
ext_mem_alloc_type_aligned (pmm.s2.c: 600)
```
After that it calls `panic` again, closing the loop.

**Additional context**
I was trying out Limine on a P6-class x86Box machine and ran into a boot loop. It may be that this machine doesn't support 0xE820 and that Limine reads garbage and passes it to the PMM. But regardless what causes the PMM initialization to fail, it calls panic, which decides to initialize textmode, which causes a framebuffer to be allocated, which fails due to the PMM not being initialized yet.

The core problem being reported here is thus a circular dependency which causes and infinite loop if Stage 2 panics too early.

Mintsuki:
Fixed.

PR #142, opened 2021-12-10
Made panic more robust for BIOS

lukflug:
See issue #140.
* Added makefile recipes to replicate issue more easily and test on legacy machines, on a DOS-readable image.
* Added crude print routine based on `int 10h, 0x0E`, in case stage 2 panic detects a recursive invocation.
* Made VGA textmode driver write directly to framebuffer, instead triple buffering, if the PMM does not allow allocations.

The VGA textmode driver fix will create artifacts, but it will only be invoked when a panic happens before the PMM is done initializing anyways, and it is enough to produce a readable panic screen.
The print routine in panic itself serves as a last line of defense, in case there is another circular dependency somewhere and does not produce a stack trace.

### DIFF
--- a/stage23/drivers/vga_textmode.s2.c
+++ b/stage23/drivers/vga_textmode.s2.c
@@ -139,7 +139,7 @@ void init_vga_textmode(size_t *_rows, size_t *_cols, bool managed) {
         current_video_mode = -1;
     }
 
-    if (!pmm_allocation_allowed()) {
+    if (allocations_disallowed) {
         front_buffer = back_buffer = video_mem;
     }
 
--- a/stage23/drivers/vga_textmode.s2.c
+++ b/stage23/drivers/vga_textmode.s2.c
@@ -139,10 +139,6 @@ void init_vga_textmode(size_t *_rows, size_t *_cols, bool managed) {
         current_video_mode = -1;
     }
 
-    if (allocations_disallowed) {
-        front_buffer = back_buffer = video_mem;
-    }
-
     if (back_buffer == NULL) {
         back_buffer = ext_mem_alloc(VD_ROWS * VD_COLS);
     } else {
--- a/stage23/drivers/vga_textmode.s2.c
+++ b/stage23/drivers/vga_textmode.s2.c
@@ -139,6 +139,10 @@ void init_vga_textmode(size_t *_rows, size_t *_cols, bool managed) {
         current_video_mode = -1;
     }
 
+    if (!pmm_allocation_allowed()) {
+        front_buffer = back_buffer = video_mem;
+    }
+
     if (back_buffer == NULL) {
         back_buffer = ext_mem_alloc(VD_ROWS * VD_COLS);
     } else {
--- a/stage23/lib/panic.s2.c
+++ b/stage23/lib/panic.s2.c
@@ -12,6 +12,7 @@
 
 static const char* recursive_panic = 0;
 
+#if bios == 1
 void fallback_print (const char* string) {
     int i;
     struct rm_regs r = {0};
@@ -20,6 +21,7 @@ void fallback_print (const char* string) {
         rm_int(0x10,&r,&r);
     }
 }
+#endif
 
 __attribute__((noreturn)) void panic(const char *fmt, ...) {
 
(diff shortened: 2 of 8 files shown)


stage23/mm/pmm.s2.c
```diff
@@ -801,3 +801,7 @@ bool memmap_alloc_range(uint64_t base, uint64_t length, uint32_t type, bool free
 
     return false;
 }
+
+bool pmm_allocation_allowed(void) {
```
Mintsuki: We don't do getters and setters, just check the `allocations_disallowed` variable directly.

stage23/mm/pmm.h
```diff
@@ -57,4 +57,6 @@ void pmm_reclaim_uefi_mem(void);
 void pmm_release_uefi_mem(void);
 #endif
 
+bool pmm_allocation_allowed();
```
Mintsuki: See comment in pmm.s2.c; also no `(void)`

stage23/lib/panic.s2.c
```diff
@@ -10,13 +10,58 @@
...
     va_list args;
 
     va_start(args, fmt);
 
     quiet = false;
 
+    if (recursive_panic) {
+#if bios == 1
+        struct rm_regs r = {0};
+        rm_int(0x11,&r,&r);
```
Mintsuki: Coding style: space after comma in arguments

stage23/lib/panic.s2.c
```diff
@@ -10,13 +10,58 @@
...
+        switch ((r.eax>>4)&3) {
```
Mintsuki: Coding style: space between operators

stage23/lib/panic.s2.c
```diff
@@ -10,13 +10,58 @@
...
 #include <mm/pmm.h>
 
+static const char* recursive_panic = 0;
+
+#if bios == 1
+void fallback_print (const char* string) {
+    int i;
+    struct rm_regs r = {0};
+    for (i=0; string[i]; i++) {
+        r.eax = 0x0E00|string[i];
```
Mintsuki: Coding style: hex should be lowercase, space between operators.

stage23/lib/panic.s2.c
```diff
@@ -10,13 +10,58 @@
 #include <lib/term.h>
 #include <mm/pmm.h>
 
+static const char* recursive_panic = 0;
+
+#if bios == 1
+void fallback_print (const char* string) {
```
Mintsuki: Coding style: No space after function name, use right side pointers.

stage23/drivers/vga_textmode.s2.c
```diff
@@ -139,6 +139,10 @@ void init_vga_textmode(size_t *_rows, size_t *_cols, bool managed) {
         current_video_mode = -1;
     }
 
+    if (!pmm_allocation_allowed()) {
```
Mintsuki: Address bool variable directly, see pmm.s2.c

stage23/drivers/vga_textmode.s2.c
```diff
@@ -139,6 +139,10 @@ void init_vga_textmode(size_t *_rows, size_t *_cols, bool managed) {
...
+        front_buffer = back_buffer = video_mem;
```
Mintsuki: This could cause issues, I'd just do this check in `term_textmode()` and `term_vbe()` so if allocations are not allowed, the terminal gets always set to not ready.

Mintsuki (changes requested):

stage23/drivers/vga_textmode.s2.c
lukflug: Wouldn't setting it to not ready cause nothing to be printed out?
It is obviously a very crude thing to do and does cause artifacts on the screen, but it is fine for panic.
What about setting it to not ready, unless the recursive panic flag is set, such that this never gets invoked under any other situation?
image

Mintsuki: Then a better idea would be, in `panic()` to, instead of `if (recursive)`, do `if (allocations_disallowed && term_backend == NOT_READY)` and avoid the recursive panic entirely anyways, since it can only really happen this way.

Mintsuki:
LGTM

PR #164, opened 2022-03-21
Ignore spaces at the start of lines in the config file

benjidial:
This allows lines to be indented. I made this because I wanted to indent the local keys in every entry of my config file, to make it more readable.


Mintsuki:
Hi, LGTM but you should also consider TAB characters as whitespace.

PR #169, opened 2022-04-28
Specify the starting index of revision numbers

Noodlez1232:
When I was reading this originally, I was a bit confused about what to set the revision numbers, as no "lowest version" was specified. I was only able to find this out by looking at the barebones example.

I think it might also be worth specifying revisions for each structure provided in the protocol specification, so that there is no confusion there.


Mintsuki:
LGTM

PR #173, opened 2022-05-26
Elf sections: Ensure every section is loaded and set its address.

NathanRoyer:
This is in conformance of the Multiboot2 specification 2.0, paragraph 3.6.7.


Mintsuki:
Fixed upstream in ab79b95

PR #205, opened 2022-08-17
Initial AArch64 port

qookei:
Tested on:
 - QEMU TCG (`-M virt -cpu cortex-a72`),
 - QEMU KVM (`-M virt -cpu host` on a RPi4),
 - RPi4 bare metal with [UEFI from PFTF](https://github.com/pftf/RPi4).

What works:
 - All the generic stuff (terminal, menu, disk and keyboard I/O, etc.),
 - EFI chainloading,
 - Limine boot protocol (incl. SMP and terminal runtime).

What's left to do:
 - Device tree support (currently only ACPI is supported),
 - ACPI parking protocol for SMP core bring-up,
 - Linux boot protocol (not strictly necessary, and of limited use as it requires a device tree, and EFISTUB is a thing),
 - Bug fixes for U-Boot,
 - TODOs for things I can't test (5-level paging and HW random number generator).


PROTOCOL.md
```diff
@@ -155,6 +156,46 @@ to the stack before jumping to the kernel.
 
 All other general purpose registers are set to 0.
 
+### aarch64
+
+`PC` will be the entry point as defined as part of the executable file format,
+unless the an Entry Point feature is requested (see below), in which case,
+the value of `PC` is going to be taken from there.
+
+The contents of the `VBAR_EL1` register are undefined, and the kernel must load it's own.
```
Dennisbonke, line 165: Nitpick, the kernel must load it's own is wrong, the kernel must load its own is better. We don't mean load it is own here.

Dennisbonke (changes requested):
I spotted one issue in the protocol.md file. I've added a suggestion on how to fix it. For the code I'm afraid I don't know enough about AArch64.

Andy-Python-Programmer:
Amazing! 🥳

PR #236, opened 2022-11-18
Fixed little memory leak in the UEFI build

beakeyz:
this pr fixes a bug that would make the bootloader crash on systems with larger disks.
this is because when scanning for partitions, it would keep allocating volume structures, even when it didn't find any partition, and not clean them up.


Mintsuki:
LGTM thanks!

PR #240, opened 2022-12-06
adding bytes description to stack_size

TornaxO7:


TornaxO7:
what, I only pulled from trunk ;-;

PR #252, opened 2023-03-07
limine: Add an arch tag to simplify multi-arch netboot

blanham:
As I mentioned on the Discord, there's no easy way to automatically netboot multiple architectures from one PXE server, as there's no way to specify per-arch tftp root directories (at least with dnsmasq's bootp/tftp implementation), which would limit one to a singular `limine.cfg`. This PR adds an ARCH tag (with KERNEL_ARCH as an alias) so that limine can filter out incompatible kernels.

### DIFF
--- a/common/lib/misc.c
+++ b/common/lib/misc.c
@@ -21,6 +21,20 @@ UINT32 efi_desc_ver = 0;
 
 bool editor_enabled = true;
 
+const char *get_arch_name(void) {
+#ifdef __i386__
+    return "i386";
+#elif defined (__x86_64__)
+    return "x86_64";
+#elif defined (__aarch64__)
+    return "aarch64";
+#elif defined (__riscv64)
+    return "riscv64";
+#else
+    #error "Unknown architecture"
+#endif
+}
+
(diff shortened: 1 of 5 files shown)


common/lib/misc.c
```diff
@@ -21,6 +21,18 @@ UINT32 efi_desc_ver = 0;
...
 
+const char *get_arch_name(void) {
+#ifdef __i386__
+    return "i386";
+#elif defined (__x86_64__)
+    return "x86_64";
+#elif defined (__aarch64__)
+    return "aarch64";
+#else
+    return "unknown";
```
48cf: I feel like this should be an `#error` just to force anyone porting Limine to a new architecture to handle it

common/lib/misc.c
```diff
@@ -21,6 +21,18 @@ UINT32 efi_desc_ver = 0;
 
 bool editor_enabled = true;
 
+const char *get_arch_name(void) {
+#ifdef __i386__
```
48cf, line 25: Change `#ifdef` to `#if defined` so it's consistent with the rest of the code

Mintsuki:
This is superseded by the `${ARCH}` config macro.

PR #253, opened 2023-03-07
Add a note about the kernel being physically contignous

pitust:


Mintsuki:
Fixed in ca694756

Andy-Python-Programmer:
:trollface:

PR #261, opened 2023-04-10
limine: Fix the order of fields in `struct limine_video_mode`

sleepy-monax:
The order of the fields `pitch`, `width`, and `height` is inconsistent between `struct limine_video_mode` and `struct limine_framebuffer`


Mintsuki:
Breaks ABI.

sleepy-monax:
This is why we can't have nice things

Andy-Python-Programmer:
lmfao

PR #263, opened 2023-04-12
multiboot2: Support TEXTMODE config as with linux

klange:
Cribbing off the implementation in `common/protos/linux.c`, this adds support for the `TEXTMODE` config variable with Multiboot2. This is useful if a kernel provides a framebuffer header tag with a preferred mode (preventing text mode from being enabled unless no framebuffer can be configured), but a user would like to use text mode anyway (perhaps for debugging, or out of a sense of nostalgia).

As with the existing implementation for Linux, this config value is only checked in BIOS configurations - no attempt is made to examine it under EFI or report an error in that case.

### DIFF
--- a/common/protos/multiboot2.c
+++ b/common/protos/multiboot2.c
@@ -521,6 +521,15 @@ noreturn void multiboot2_load(char *config, char* cmdline) {
         size_t req_width = 0;
         size_t req_height = 0;
         size_t req_bpp = 0;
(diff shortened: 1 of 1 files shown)


Andy-Python-Programmer (approved):

Mintsuki:
This should also be documented in `CONFIG.md`.

Mintsuki:
I'll do it myself.

PR #264, opened 2023-04-12
add warning text

SachaTending:
sorry for bad english


lukflug:
I think that the language in text visible to the end user is the most critical thing to get right.

PR #267, opened 2023-04-24
Implement Multiboot2 network tag

blanham:
Added caching of the dhcp response in the pxe code, and then adding the tag containing it.

### DIFF
--- a/common/pxe/tftp.s2.c
+++ b/common/pxe/tftp.s2.c
@@ -10,6 +10,10 @@
 #include <mm/pmm.h>
 #include <lib/misc.h>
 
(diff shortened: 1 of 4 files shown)


common/pxe/tftp.s2.c
```diff
@@ -10,13 +10,19 @@
 #include <mm/pmm.h>
 #include <lib/misc.h>
 
+// cache the dhcp packet
+uint8_t cached_dhcp_packet[1472] = { 0 };
```
Andy-Python-Programmer, line 14: Might want to put the packet size as a constant (magic values :ghost:)?

PR #268, opened 2023-04-26
Implement UEFI serial protocol and add baudrate configuration

blanham:
This is implemented in such a way that it will fallback to the normal serial port if the serial io protocol is not found. Baudrate still defaults to 9660. I also included a note about potentially extending serial support to devices with 16550 style UARTs, probably by adding config options to specify an MMIO base and register width. I also moved the serial boolean to the driver, I think it makes more sense there than in misc.c. Finally, I added an entry to .gitignore for my future convenience, I use a tool called compiledb to generate said file to improve my experience using VSCode

I realized I neglected to change readline to use serial on UEFI systems. The additions are in the second commit, it's not the cleanest implementation in the world and I'm open to suggestions to improve it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /common/flanterm
 /common/stb/stb_image.h
 /ovmf*
+compile_commands.json
 *.o
 *.d
 *.a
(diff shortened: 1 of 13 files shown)


Mintsuki:
The UEFI part of this PR is problematic as I explained a while ago on Discord. The BIOS part is fine, but that would make this setting BIOS only...

PR #274, opened 2023-05-29
Initial riscv64 port

xvanc:
Adds support for RISC-V (rv64imac) with UEFI.

Tested on QEMU (`-machine virt -cpu rv64`). Limine protocol and EFI chainloading work. Linux boot should be a relatively trivial addition, but is poorly documented and I want to be sure I fully understand it before implementing it.

Currently ACPI is required to detect other harts for SMP, which is less than ideal since the current state of ACPI on RISC-V is practically unusable. If support were added for parsing device trees this requirement could be removed. (Linux boot may also require this.)


Andy-Python-Programmer:
🥳

PROTOCOL.md
```diff
@@ -632,6 +658,78 @@ struct limine_video_mode {
...
+#define LIMINE_PAGING_MODE_RISCV_SV39 0
+#define LIMINE_PAGING_MODE_RISCV_SV48 1
+#define LIMINE_PAGING_MODE_RISCV_SV57 2
+```
+
+No `flags` are currently defined.
+
+The default mode (when this request is not provided) is `LIMINE_PAGING_MODE_RISCV_SV39`.
+
 ### 5-Level Paging Feature
```
Andy-Python-Programmer, line 745: The "5-Level Paging Feature" should have a note mentioning that it has been deprecated.

```suggestion
### 5-Level Paging Feature
*Note: This feature is deprecated, use the "Paging Mode Feature" instead.*
```

PROTOCOL.md
```diff
@@ -197,6 +197,32 @@ Size Request (see below).
 
 All other general purpose registers (including `X29` and `X30`) are set to 0. Vector registers are in an undefined state.
 
+### riscv64
```
DeanoBurrito, line 200: Pendantic, but it's probably worth mentioning that EFI boot services are exited? (like the other platforms)

PROTOCOL.md
```diff
@@ -197,6 +197,32 @@ Size Request (see below).
...
+
+`x1`(`ra`) is undefined, the kernel must not return from the entry point.
+
+`x2`(`sp`) is set to point to a stack, in bootloader-reclaimable memory, which is
+at least 64KiB (65536 bytes) in size, or the size specified in the Stack
+Size Request (see below).
+
+`x3`(`gp`) is undefined, kernel must load it's own global pointer if needed.
+
+All other general purpose registers, with the exception of `xX`(`t0`), are set to 0.
```
DeanoBurrito: `xX` should probably be `x5`

DeanoBurrito (commented):
The field names of `limine_smp_response` and `limine_smp_info` are also different for riscv but not updated in the protocol docs.

common/lib/mem.asm_riscv64
```diff
@@ -0,0 +1,58 @@
...
+        sb      t1, (t0)
+        addi    a2, a2, -1
+        bnez    a2, 0b
+1:      ret
+
+
+.global memcmp
+memcmp:
+        mv      t0, a0
+        beqz    a2, 1f
```
klange: `memcmp` should return `0` when `length` is `0`, looks like we're getting the first address out here.

test.mk
```diff
@@ -236,6 +240,30 @@ uefi-aa64-test:
...
+	sudo partprobe `cat loopback_dev`
+	sudo mkfs.fat -F 32 `cat loopback_dev`p1
+	sudo mount `cat loopback_dev`p1 test_image
+	sudo mkdir test_image/boot
+	sudo cp -rv $(BINDIR)/* test_image/boot/
+	sudo cp -rv test/* test_image/boot/
+	sudo $(MKDIR_P) test_image/EFI/BOOT
+	sudo cp $(BINDIR)/BOOTRISCV64.EFI test_image/EFI/BOOT/
+	sync
+	sudo unmount test_image/
```
klange: `umount`

test.mk
```diff
@@ -236,6 +240,30 @@ uefi-aa64-test:
...
+	sudo losetup -d `cat loopback_dev`
+	rm -rf test_image loopback_dev
+	qemu-system-riscv64 -m 512M -M virt -cpu rv64 -drive if=pflash,unit=1,format=raw,file=ovmf-rv64/OVMF.fd -net none -smp 4 -device ramfb -device qemu-xhci -device usb-kbd -hda test.hdd -serial stdio
```
klange: I'm on a very recent git build of qemu and `-hda` doesn't seem to produce a working storage device.

Long-form virtio block device was recommend to me and works: `-drive file=test.hdd,format=raw,id=hd0 -device virtio-blk-device,drive=hd0`

common/GNUmakefile
```diff
@@ -125,6 +129,26 @@ ifeq ($(TARGET),uefi-aarch64)
         -DUEFI
 endif
 
+ifeq ($(TARGET),uefi-riscv64)
+	ifeq ($(CC_FOR_TARGET_IS_CLANG),yes)
+		override RISCV_CFLAGS += -march=rv64imac -mabi=lp64
+	else
+		override RISCV_CFLAGS += -march=rv64imac_zicsr_zifencei -mabi=lp64
```
klange: I'm not sure where the fault lies, but building with this ABI on gcc 11.3 as packaged for https://packages.ubuntu.com/jammy/gcc-riscv64-linux-gnu is getting me the "soft float" ABI which is incompatible with the "double-float" ABI that mintsuki's libgcc binaries were built for, and binutils ld 2.38 is refusing to link the two halves.

As changing it on this end was easier, I switched to an march/mabi with `d` support and got a working build out of it, but I'm not sure who's wrong here.

test.mk
```diff
@@ -236,6 +240,30 @@ uefi-aa64-test:
 	rm -rf test_image loopback_dev
 	qemu-system-aarch64 -m 512M -M virt -cpu cortex-a72 -bios ovmf-aa64/OVMF.fd -net none -smp 4 -device ramfb -device qemu-xhci -device usb-kbd  -hda test.hdd -serial stdio
 
+.PHONY: uefi-rv64-test
+uefi-rv64-test:
+	$(MAKE) ovmf-rv64
+	$(MAKE) test-clean
+	$(MAKE) test.hdd
+	$(MAKE) limine-uefi-riscv64
+	$(MAKE) -C test TOOLCHAIN_FILE='$(call SHESCAPE,$(BUILDDIR))/toolchain-files/uefi-riscv64-toolchain.mk'
```
klange, line 249: These `test` targets are apparently broken outside of x86-64, but I commented this out to test the rest of the rule with no payload kernels.

test.mk
```diff
@@ -11,6 +11,10 @@ ovmf-aa64:
 	mkdir -p ovmf-aa64
 	cd ovmf-aa64 && curl -o OVMF-AA64.zip https://efi.akeo.ie/OVMF/OVMF-AA64.zip && 7z x OVMF-AA64.zip
 
+ovmf-rv64:
+	mkdir -p ovmf-rv64
+	cd ovmf-rv64 && curl -o OVMF.fd https://retrage.github.io/edk2-nightly/bin/RELEASERISCV64_VIRT.fd
```
klange: Should this be expanding the firmware file to 33554432 bytes or whatever? I guess that may be a moving target depending on the QEMU version...

common/lib/mem.asm_riscv64
klange: Obviated by 679fdd4351e44a067a2b3d80f8d5f4447a58d5a4

common/GNUmakefile
klange: ABI for the libgcc builds was changed in https://github.com/mintsuki/libgcc-binaries/commit/8398d4287aacee886959669e284ca4bf8aad5fb7

test.mk
xvanc: This *should* now work with the latest commit.

Andy-Python-Programmer (approved):

PR #291, opened 2023-08-04
Add option to hide the help text

sophuric:
I think this option would be nice for people who prefer minimalism.

screenshot


Mintsuki:
Please change this PR to target trunk rather than v5.x-branch, but other than that, it LGTM, thanks!

PR #331, opened 2024-01-31
Multiboot2 improvements

phip1611:
This improves the Multiboot2 code:

- fixes a bug to align the behavior with GRUB
- better logging of unknown Multiboot header tags

**Please** look into the commit messages for further explanation.

I didn't test it as I'm still figuring out how the test infrastructure works.

### DIFF
--- a/common/protos/multiboot2.c
+++ b/common/protos/multiboot2.c
@@ -214,8 +214,17 @@ noreturn void multiboot2_load(char *config, char* cmdline) {
             }
 
             default:
-                if (is_required)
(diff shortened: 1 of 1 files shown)


Mintsuki:
Thank you for the PR.

The test setup I usually use is to first `./bootstrap` of course, then `./configure --enable-all`, and then run `make clean test-clean fat32-test` or `make clean test-clean uefi-x86-64-test` depending on BIOS vs UEFI (or one of the many other test targets).

Issue #330, opened 2024-01-31
Onboard to nixpkgs


phip1611:
Hi, cool project! 
It would be awesome to upstream this to [nixpkgs](https://github.com/NixOS/nixpkgs)! My company and I do a lot of low-level OS-related kernel development with the help of [Nix](https://nixos.org/). Having everything packaged in Nix enables us to create reproducible and productive development environments.

As this is my last day of work before my long vacation starts, I don't want to upstream this to nixpkgs today. I'm happy to work towards that goal when I'm back in March!

For now, I'd like to share my progress so that anyone else can continue working on this - if interested. Currently, I package it in Nix like that, but without the cross-compilation for riscv64 and aarch64:

```nix
# Builds limine with all available features.

{ gcc, gnumake, mtools, nasm, stdenv }:

let
  # It is recommended by limine to use the release tarballs instead the git repo.
  limineUrl = "https://github.com/limine-bootloader/limine/releases/download/v7.0.2/limine-7.0.2.tar.gz";
  limineSrc = builtins.fetchTarball {
    url = limineUrl;
    sha256 = "sha256:0h70hkzyxpavsa896mj8pqc4gr6hj0bb2d76prp0rsgy1syyy9kc";
  };
in
stdenv.mkDerivation {
  # TODO license meta data
  pname = "limine";
  version = "7.0.2";
  src = limineSrc;
  nativeBuildInputs = [ mtools gnumake gcc nasm ];
  doCheck = false;
  # TODO missing --enable-uefi-{aarch64,riscv64} cross compilation so far
  configurePhase = ''
    ./configure \
      --enable-bios-cd \
      --enable-bios-pc \
      --enable-bios-pxe \
      --enable-uefi-ia32 \
      --enable-uefi-x86-64 \
      --enable-uefi-cd
  '';
  buildPhase = ''
    make -j $(nproc)
  '';
  installPhase = ''
    mkdir -p $out/bin
    cp -R bin/. $out
    ln -s $out/limine $out/bin/limine
  '';
}
```

ping @blitz

AshesOfEther:
This was already tried at NixOS/nixpkgs#194841, but it was never reviewed by the relevant maintainers

phip1611:
Oh, good to know! Yeah, the throughput in nixpkgs on GitHub is incredible and maintainers are all overwhelmed.

My colleques and I know a few maintainers personally. I'm confident that we can do something about that issue.

AshesOfEther:
[...]

PR #332, opened 2024-01-31
repo: init flake.nix

phip1611:
First of all: This is a proposal. I don't expect you to merge this. But I'd like to highlight the benefits and then you can decide 😄 

This inits a flake.nix. Nix users will get a shell with all relevant tools to build limine this way. The only missing bit is that `./configure --enable-uefi-riscv64 --enable-uefi-aarch64`
won't work as I didn't manage to get a proper cross-compiler in the Nix environment. I tried adding `pkgsCross.aarch64-multiplatform.gcc` to the shell but the downside is that this is not cached in cache.nixos.org; building takes ages..

I also tried switching to clang instead of gcc, but `./configure` still reported that there is no cross-toolchain.

Anyway: I think this is a improvement to the current situation.

PS: In case you are not familiar with Nix/NixOS: This here is completely orthogonal to https://github.com/limine-bootloader/limine/issues/330. It is just about developer experience in this repo.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,48 @@
+{
+  "nodes": {
+    "flake-parts": {
+      "inputs": {
+        "nixpkgs-lib": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1706569497,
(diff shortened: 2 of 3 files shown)


Mintsuki:
This PR was closed unintentionally due to branch renaming. Please reopen it if you intend on continuing work on this.

phip1611:
Hi, I'm back from vacation and ready to work on that PR now!

PR #339, opened 2024-03-19
nix: init flake.nix

phip1611:
**TL;DR**: This boosts the productivity of all people using NixOS or Nix in general in their environment. It provides you with the right environment to build this project.

This inits a flake.nix into the repository that exports a convenient Nix shell. When using `$ nix develop`, users will get a shell with all relevant tools to build the repository. If one uses a "direnv" integration into their shell, the environment is automatically loaded.

Not directly #330 but related to it.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,57 @@
+# This flake is supposed to enable a convenient development environment.
+# It is independent of any packaging in nixpkgs.
+#
+# See https://github.com/limine-bootloader/limine/issues/330 for more details
+# regarding the packaging in nixpkgs.
+
+{
+  description = "limeboot";
+
+  inputs = {
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    flake-parts.inputs.nixpkgs-lib.follows = "nixpkgs";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
+  };
+
+  outputs = inputs@{ self, nixpkgs, flake-parts }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      flake = { };
+      # Don't artificially limit users at this point. If a build fails, they
+      # will notice it soon enough.
+      systems = nixpkgs.lib.systems.flakeExposed;
+      perSystem = { config, pkgs, ... }:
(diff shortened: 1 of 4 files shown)


phip1611:
It will not cause any maintenance burden for you. For example, it is perfectly fine to not update the flake inputs for a year or two. Nix(OS) users however will be very happy about this.

PS: You can ping me from time to time if you need help (assuming that you are not familiar with Nix yet - correct me if I'm wrong)

flake.nix
```diff
@@ -0,0 +1,57 @@
+# This flake is supposed to enable a convenient development environment.
+# It is independent of any packaging in nixpkgs.
+#
+# See https://github.com/limine-bootloader/limine/issues/330 for more details
+# regarding the packaging in nixpkgs.
+
+{
+  description = "limeboot";
```
AshesOfEther, line 8: Limeboot?
```suggestion
  description = "limine";
```

flake.nix
```diff
@@ -0,0 +1,57 @@
...
+          devShells = {
+            default = pkgs.mkShell {
+              packages = with pkgs; [
+                # Dependencies for ./bootstrap
+                autoconf
+                automake
+
+                # General build dependencies
+                cacert
+                git
```
AshesOfEther, line 33: I think we also want Make
```suggestion
                git
                gnumake
```

AshesOfEther (changes requested):
Few things, otherwise LGTM!

flake.nix
phip1611: Yikes, thanks. This is the part that I didn't check a second time. 😂

AshesOfEther:
GitHub is weird so I ended up marking my review as duplicate, woops :sweat_smile:

flake.nix
JohnAZoidberg: Just FYI: It's provided by default by nix, same as gcc. See below.

Mintsuki:
Hi, thanks for the PR!

I would say I am familiar enough with nix already :p

phip1611:
Thanks for merging, nice!

PR #340, opened 2024-03-21
nix: build limine in Nix derivation + CI

phip1611:
**TL;DR**: `nix build .\#limine` now builds limine. All relevant files are then in `./result`.

This was a tough challenge due to the nature of the project. The self-hacked
"git submodule checkout with additional patches" in `./bootstrap` is very 
non-standard. But it should work now. 

The big advantage of packaging Limine in Nix directly in this repo is, instead
of consuming it from nixpkgs, one can quickly change sources and see results 
while still being able to have the whole project packaged in Nix.

This was fun :D

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -29,3 +29,21 @@ jobs:
 
       - name: Build the bootloader (GNU, riscv64)
         run: ./bootstrap && ./configure TOOLCHAIN_FOR_TARGET=riscv64-linux-gnu --enable-werror --enable-uefi-riscv64 && make all && make maintainer-clean
+
+  build_nix:
+    name: Build with Nix
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: cachix/install-nix-action@v26
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+      - run: nix build .#limine
+
+  build_nix_shell:
+    name: Build with Nix shell toolchain
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: cachix/install-nix-action@v26
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+      - run: nix develop --command bash -c "./bootstrap && ./configure --enable-all && make -j $(nproc)"
--- a/nix/build.nix
+++ b/nix/build.nix
@@ -14,6 +14,7 @@
   # Helpers
   fd
 , lib
+, keep-directory-diff
 , nix-gitignore
 , stdenv
 
@@ -28,33 +29,42 @@
 }:
 
 let
+  # The actual current source but close to the repo state, i.e., no files from
+  # the bootstrap step.
   currentRepoSrc = nix-gitignore.gitignoreSource [
     # Additional git ignores:
     "flake.nix" # otherwise
     "flake.lock"
     "nix/"
+
+    # bootstrapped sources from ./bootstrap
+    "/build-aux/freestanding-toolchain"
+    "/freestanding-headers"
+    "/decompressor/cc-runtime"
+    "/limine-efi"
+    "/tinf"
+    "/common/flanterm"
+    "/stb"
   ] ../.;
 
   # Contains the sources downloaded by the Git submodule-like initialation done
   # in ./bootstrap.
   #
   # ALWAYS update the hash when one of the network dependencies in ./bootstrap
-  # changes. Also, before updating, it is recommended to run "make clean"
-  # beforehand.
+  # changes.
   # bootstrappedSrcHash = lib.fakeHash;
-  #
-  # TODO: Unfortunately, currently this hash changes for almost every repository
-  # change. We need to strip down this derivation further to only contain the
-  # changed sources.
-  bootstrappedSrcHash = "sha256-UU5pkdbaKXPs/i/hnuk4vZcxiag1cTsTCcn2LGzPuMs=";
-  bootstrappedSrc = stdenvNoCC.mkDerivation {
-    pname = "limine-bootstrapped";
+  bootstrappedSrcHash = "sha256-uggy1cDftq0tPD777hS+rz2oBnP2Q2AQ9xHdM8QhBQQ=";
+
+  # The full accumulated source tree to build Limine from.
+  bootstrappedSrc = stdenv.mkDerivation {
+    pname = "limine-src-bootstrapped";
     version = "0.0.0";
     src = currentRepoSrc;
     nativeBuildInputs = [
       cacert
-      git
       fd
+      git
+      keep-directory-diff
     ];
     buildPhase = ''
       runHook preBuild
@@ -62,6 +72,8 @@ let
       # `true` refers to the binary/bash-builtin to prevent any configuration
       # steps apart from downloading sources.
       AUTOMAKE=true AUTORECONF=true ./bootstrap
+
+      keep-directory-diff ${currentRepoSrc} .
 
       # When cloning, Git automatically creates hooks. Unfortunately, in a Nix
       # environment / on a NixOS system, this includes Nix store paths.
--- a/nix/build.nix
+++ b/nix/build.nix
@@ -1,0 +1,139 @@
+# Building Limine with all features in Nix.
+#
+# Independent of the packing in nixpkgs, This is convenient for prototyping and
+# local development.
+#
+# These derivations always builds Limine from the current src tree and not some
+# stable release. Further, unlike the nixpkgs derivation, this derivation runs
+# the ./bootstrap step which needs network access. Due to the nature of the
+# self-hacked Git submodules download approach, packaging this project in Nix
+# is especailly complicated. The complicated multi-derivation approach below
+# is the best I can get after multiple hours of trying. :).
+
+{
+  # Helpers
+  fd
+, lib
+, nix-gitignore
+, stdenv
+
+  # Actual derivation dependencies.
+, autoconf
+, automake
+, cacert
+, git
+, llvmPackages
+, mtools
+, nasm
+}:
+
+let
+  currentRepoSrc = nix-gitignore.gitignoreSource [
+    # Additional git ignores:
+    "flake.nix" # otherwise
+    "flake.lock"
+    "nix/"
+  ] ../.;
+
+  # Contains the sources downloaded by the Git submodule-like initialation done
+  # in ./bootstrap.
+  #
+  # ALWAYS update the hash when one of the network dependencies in ./bootstrap
+  # changes. Also, before updating, it is recommended to run "make clean"
+  # beforehand.
+  # bootstrappedSrcHash = lib.fakeHash;
+  #
+  # TODO: Unfortunately, currently this hash changes for almost every repository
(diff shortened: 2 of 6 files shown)


phip1611:
In case you need help maintaining this in the future, feel free to ping me.

PS: We should still work towards upstreaming this to nixpkgs in the next and last step. I think I can work on it in the upcoming week

---

If this repository would use regular git submodules and apply the patches in a dedicated step, the packaging would be much easier. 

Is there a reason why this is not done? 

---

**Update** It works! This was for sure the most complex thing I ever packaged in Nix. 


Packaging it in nixpkgs will be much easier luckily, as when we download the release tarball, we do not have to run the `./bootstrap` script.

phip1611:
Yikes, CI failure.... packaging the stuff that `./bootstrap` does it very complicated. I already invested a few hours into it. I have a couple of more ideas. But maybe we have to give up here...Hopefully not!

phip1611:
If this repository would use regular git submodules and apply the patches in a dedicated step, the packaging would be much easier. 

Is there a reason why this is not done?

nix/build.nix
```diff
@@ -0,0 +1,191 @@
...
+
+      make all -j $(nproc)
+
+      runHook postBuild
+    '';
+    installPhase = ''
+      runHook preInstall
+
+      mkdir -p $out/bin
+      cp -R bin/. $out
```
Mintsuki: Is there a reason that `DESTDIR=xyz make install` is not used here? Copying files manually out of `bin` as a method to perform an install isn't quite supported nor expected.

Furthermore, the command `limine --print-datadir` is hardcoded at build time to point to the datadir (aka usually `/usr/share/limine`) -- I know that nix is a bit odd when it comes to paths so I am not sure how that will play out.

nix/build.nix
```diff
@@ -0,0 +1,191 @@
...
+              echo "$filename was created by ./bootstrap"
+          fi
+      done
+
+      # This should report nothing. Othewise, the Nix build will fail.
+      # grep -r /nix/store .
+
+      runHook postBuild
+    '';
+    dontPatchShebangs = true; # probalby unneeded
```
Mintsuki: Spelling.

bootstrap
```diff
@@ -1,4 +1,4 @@
-#! /bin/sh
+#!/usr/bin/env bash
```
Mintsuki: We're not going to change scripts to hard depend on `bash`. I would suggest keeping the shebang as is, and add variables for `AUTORECONF` and `AUTOMAKE` that default to `autoreconf` and `automake` respectively. Then, when calling `./bootstrap`, one can pass `AUTORECONF=true` to skip the step.

.editorconfig
```diff
@@ -14,5 +14,5 @@ max_line_length = 80
 [{Makefile,GNUmakefile*}]
 indent_style = tab
 
-[{*.nix,*.yml}]
+[{*.nix,*.sh,*.yml}]
```
Mintsuki: The indent style for *.sh should be 4.

Mintsuki:
> Is there a reason why this is not done?

It could probably be done using git submodules, it's just that I feel more comfortable working with this bootstrapping method, at least for now. I never really quite liked git submodules.

bootstrap
phip1611: Oh sorry. This was just for testing. I can revert it

phip1611:
>  I never really quite liked git submodules.

nobody really does :)

nix/build.nix
phip1611: ahh, good point! The same probably applies for https://github.com/NixOS/nixpkgs/pull/298048

I think I can fix it/improve it

bootstrap
phip1611: got it, good idea

phip1611:
Ping @mintsuki - are we good to go here? I'm confident that the derivation is now in a good, productive, and ready to use shape.

btw. To see why this is useful: One can create bootable images in Nix using limine, as I did in here: https://github.com/phip1611/nixos-configs/pull/54/files

bootstrap
phip1611: Yikes, I forgot this. Sorry. Will do it tomorrow @mintsuki

.github/workflows/check.yml
```diff
@@ -29,3 +29,21 @@ jobs:
 
       - name: Build the bootloader (GNU, riscv64)
         run: ./bootstrap && ./configure TOOLCHAIN_FOR_TARGET=riscv64-linux-gnu --enable-werror --enable-uefi-riscv64 && make all && make maintainer-clean
+
+  build_nix:
```
phip1611, line 33: One last thing: I think this should be set to allow failures. Otherwise, I'm concerned that changes to the repo result in frequent errors because `bootstrappedSrcHash` will be invalid.

This can eventually be solved of the todo in bootstrappedSrc is addressed

phip1611: Problem fixed. Will push soon!

bootstrap
phip1611: Addressed. I hope the solution is fine to you?

Mintsuki: There are issues with this, imho. I will shortly make a commit addressing the issue as I originally stated in this thread.

nix/build.nix
```diff
@@ -0,0 +1,153 @@
...
+# is the best I can get after multiple hours of trying. :).
+
+{
+  # Helpers
+  fd
+, lib
+, keep-directory-diff
+, nix-gitignore
+, stdenv
+, stdenvNoCC
```
Mintsuki: I don't think `stdenvNoCC` is warranted here, going by the nixpkgs PR's state.

nix/build.nix
```diff
@@ -0,0 +1,153 @@
...
+, mtools
+, nasm
+, pkg-config
+}:
+
+let
+  # The actual current source but close to the repo state, i.e., no files from
+  # the bootstrap step.
+  currentRepoSrc = nix-gitignore.gitignoreSource [
+    # Prevent that the hash ov the sources change, when we change it in Nix.
```
Mintsuki: ov -> of

nix/build.nix
```diff
@@ -0,0 +1,153 @@
...
+  # Actual derivation dependencies.
+, autoconf
+, automake
+, cacert
+, git
+, gnumake
+, llvmPackages
+, mtools
+, nasm
+, pkg-config
```
Mintsuki: Do `gnumake` and `pkg-config` need to be explicit? nixpkgs PR does not have them.

nix/build.nix
phip1611: The `bootstrappedSrc` derivation doesn't need any CC.

However, using `stdenvNoCC` instead of `stdenv` is primarily done out of purism/perfectionism and not because of a technical reason

nix/build.nix
```diff
@@ -0,0 +1,153 @@
...
+    "nix/"
+
+    # bootstrapped sources from ./bootstrap
+    "/build-aux/freestanding-toolchain"
+    "/freestanding-headers"
+    "/decompressor/cc-runtime"
+    "/limine-efi"
+    "/tinf"
+    "/common/flanterm"
+    "/stb"
```
Mintsuki, line 47: Similar to #342, shouldn't this include *all* files and directories as shown by `git status --ignored` after `./bootstrap`?

nix/build.nix
Mintsuki: This is still misspelled: probalby -> probably.

nix/build.nix
```diff
@@ -0,0 +1,153 @@
...
+  ] ../.;
+
+  # Contains the sources downloaded by the Git submodule-like initialation done
+  # in ./bootstrap.
+  #
+  # ALWAYS update the hash when one of the network dependencies in ./bootstrap
+  # changes.
+  # bootstrappedSrcHash = lib.fakeHash;
```
Mintsuki, line 55: What's this fakehash line about?

nix/build.nix
phip1611: No. This is my personal trait to be more explicit than needed, often. I remove it

nix/build.nix
```diff
@@ -0,0 +1,153 @@
...
+  bootstrappedSrcHash = "sha256-pMxjeYuNC8Yk2Qm46aCGO1UEwEiZ/RwsYypmmJQa0A4=";
+  # The full accumulated source tree to build Limine from.
+  bootstrappedSrc = stdenvNoCC.mkDerivation {
```
Mintsuki: I see now what the `stdenvNoCC` up top is needed for. But is this really necessary? Seems like an unneeded complication over just using `stdenv`.

nix/build.nix
```diff
@@ -0,0 +1,153 @@
...
+    autoconf
+    automake
+
+    gnumake
+    mtools
+    nasm
+    pkg-config # Checked for by ./configure but seems unused?!
+  ];
+in
+lib.makeOverridable stdenv.mkDerivation {
```
Mintsuki: What is this `makeOverridable` for?

phip1611: I removed it.

Background: By convention, all packages in Nix are overridable. `limine.override({ autoconf = myPatchedAutoconf })`

But I put it at the wrong place / with the current structure it is not needed.  Because `pkgs.callPackage` is used in `flake.nix` and callPackage is effectively `lib.makeOverridable` plus dependency injection, it is not needed and the package is overridable.

nix/build.nix
phip1611: Yikes. My mistake. As they are already excluded by `.gitignore`, `nix-gitignore.gitignoreSource` already takes care of that. The list are only additional entries.

nix/build.nix
phip1611: I hope this video clarifies any issues.

TL;DR: The hash must be changed when bootstrapped sources change. If the hash isn't changed, bootstrappedSrc comes from the Nix Store (if it is already there/cached).

[Screencast from 2024-04-05 13-34-28.webm](https://github.com/limine-bootloader/limine/assets/5737016/cb00e68d-6044-47c5-9568-97dac5bfac02)

bootstrap
Mintsuki: Okay, I pushed the aforementioned change. Now you can do `AUTOMAKE=true AUTORECONF=true ./bootstrap` and bootstrap will skip the autotools steps.

Likewise, if the submodules are already cloned, running `./bootstrap` will not download them again, and it will only run the autotools steps. This should remove the need to manually invoke `autoreconf`.

nix/build.nix
Mintsuki: Alright, makes sense.

bootstrap
phip1611: I rebased the branch. Thanks!

(IMHO your solution sounds unusual to me. Is this something that is widely used? Because `AUTOMAKE=true` referring to the `true` binary/bash builtin is far from intuitive. One reads it as a boolean instead.)

phip1611:
Thanks for your reviewing patience and curiosity/accuracy. Are we good to go now?

bootstrap
Mintsuki: Yeah it's a pretty standard idiom.

Mintsuki:
Yeah it LGTM now, we're good to go!

Mintsuki:
Thanks for the PR :)

PR #341, opened 2024-03-21
style: init .editorconfig

phip1611:
This adds an [`.editorconfig`](https://editorconfig.org/) file. 

In case you are not familiar with editorconfig: almost every IDE but also small terminal
editors support it (via plugins). It helps to set the right styling information as you type.

This is especially useful to have tabs in Makefiles but spaces otherwise.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+max_line_length = 80
+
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+max_line_length = 80
+
+[{Makefile,GNUmakefile.in}]
```
Mintsuki: Shouldn't this also include `GNUmakefile`?

phip1611: Fixed. I thought this is not that relevant as it is generated.

Mintsuki: The `GNUmakefile` in the root is generated, but the ones in e.g. `common/` and `decompressor/` are not.

PR #342, opened 2024-04-04
misc: init typos and github dependabot for CI workflows

phip1611:
- init <https://github.com/crate-ci/typos> and run it in CI
- add GitHub dependabot for CI workflows

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -1,0 +1,22 @@
+# Configuration for https://github.com/crate-ci/typos
+
+[files]
+extend-exclude = [
+  # Typos already uses .gitignore. These excludes are additional:
+]
+
+[default.extend-words]
+hda = "hda"
+Nd = "Nd"
+sie = "sie"
+SIE = "SIE"
+ist = "ist"
+rela = "rela"
+RELA = "RELA"
+guid = "guid"
+GUID = "GUID"
+pn = "pn"
+hd = "hd"
+htpt = "htpt"
+
+[default.extend-identifiers]
--- a/flake.nix
+++ b/flake.nix
@@ -11,9 +11,10 @@
     flake-parts.url = "github:hercules-ci/flake-parts";
     flake-parts.inputs.nixpkgs-lib.follows = "nixpkgs";
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
+    nixpkgs-unstable.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
   };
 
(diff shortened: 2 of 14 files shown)


.typos.toml
```diff
@@ -0,0 +1,28 @@
...
+extend-exclude = [
+  # Third party.
+  "build-aux/freestanding-toolchain",
+  "freestanding-headers",
+  "decompressor/cc-runtime",
+  "limine-efi",
+  "tinf",
+  "common/flanterm",
+  "stb",
+]
```
Mintsuki, line 6: Is this necessary at all considering the workflow doesn't run `./bootstrap`?

common/protos/linux.c
```diff
@@ -226,7 +226,7 @@ struct edd_device_params {
         } __attribute__ ((packed)) xprs;
         struct {
             uint64_t reserved;
-        } __attribute__ ((packed)) htpt;
+        } __attribute__ ((packed)) http;
```
Mintsuki: This is an erroneous correction.

flake.nix
```diff
@@ -46,6 +47,10 @@
 
                 # Nix
                 nixpkgs-fmt
+
+                # Misc
+                # typos is not yet frequently updated in the stable channel
+                nixpkgs-unstable.legacyPackages.${pkgs.system}.typos
```
Mintsuki, line 58: What is this for, exactly?

.typos.toml
phip1611: Yes. Imagine you run `$ typos` locally, after you ran `./bootstrap` beforehand.

common/protos/linux.c
phip1611: good catch, fixed.

flake.nix
phip1611: I'm not exactly sure what you are referring to. These are the dependencies for the Nix shell. It serves you with the environment you need for the project without needing global system state. 

image

.typos.toml
Mintsuki: Well, in that case shouldn't this include all files/directories listed by `git status --ignored` as being ignored after running `./bootstrap` (and potentially after `./configure --enable-all && make` as well)?

flake.nix
Mintsuki: Makes sense.

.typos.toml
phip1611: I updated the PR. We can remove the excludes as the are coming from `.gitignore` and typos automatically uses common ignore files

Mintsuki:
LGTM thanks!

Issue #244, opened 2023-01-09
Feature Request: aarch64 Linux protocol


yveszoundi:
I distribute a Live CD as zero configuration option for [one of my GitHub projects](https://github.com/rimerosolutions/entrusted). I've been evaluating Limine for its quality, appearance, simplicity and ability to produce painlessly small hybrid ISO images.

### Request

This is a feature request for implementing the `Linux` protocol for `aarch64`. I initially didn't read much of the documentation and I then realized at boot time that this is a `TODO` item for `aarch64`.

- It appears that the only available protocol for my use-case is `Limine`, however I'm not building "my own kernel"
- I tried extracting an ELF image from a `vmlinuz` ([vmlinuz-to-elf](https://github.com/marin-m/vmlinux-to-elf)), instead of reading more the docs...

### Rationale

For many users and/or Live CD distributors, I think that Limine can easily become the de-facto boot manager option with `aarch64` Linux protocol support.

- For an hybrid x86_64 ISO (BIOS and UEFI), Limine came as a winner for the image size (~670 MB) , followed by `Syslinux` and `Systemd-boot`  (~715 MB) and then `Syslinux` and `Grub` (~745 MB).
- I didn't need to read/understand much, prior just trying it in my Live CD build pipeline

qookei:
> It appears that the only available protocol for my use-case is `Limine`, however I'm not building "my own kernel"

Limine also supports chainloading EFI executables, which means if your kernel is compiled with EFISTUB support (`CONFIG_EFI_STUB=y`), you can chainload it from Limine, although it currently does not seem to support specifying the kernel command line (I think it should be possible to implement though).

That being said I should get around to implementing the Linux boot protocol support one day or another. Though using it comes with the caveat of not working on every machine, as it requires a device tree, and UEFI may not provide it, although there possibly is a setup option to enable it (e.g. on the RPi4), and it could be worked around by loading it from a file (but that would make the image specific to a certain SoC).

yveszoundi:
Thank you for the prompt and clear response. I'll see if I can get away with an EFI stub.

Mintsuki:
As of ac1c5d1 (latest release as well) Limine supports specifying command lines for EFI chainloading. Closing this issue for now.

craftyguy:
[...]

PR #370, opened 2024-06-09
Support Linux protocol on riscv64

ziyao233:
Initially support linux protocol on riscv64. Currently booting with UEFI + dtb is required, which is the most common case when booting on real riscv hardware (for example, u-boot emulated UEFI), since ACPI on riscv is still in draft stage, 

This could make multarch distros' work easier, as said in https://github.com/limine-bootloader/limine/issues/244#issuecomment-1747749724

### DIFF
--- a/common/lib/fdt.c
+++ b/common/lib/fdt.c
@@ -1,0 +1,35 @@
+#if !defined(__i386__) && !defined(__x86_64__)
+
+#include <stdint.h>
+#include <stddef.h>
+#include <libfdt/libfdt.h>
+
+static int fdt_get_or_add_chosen_node(void *fdt) {
+    int offset = fdt_subnode_offset(fdt, 0, "chosen");
+
+    if (offset == -FDT_ERR_NOTFOUND) {
+        offset = fdt_add_subnode(fdt, 0, "chosen");
+    }
+
+    return offset;
+}
+
+int fdt_set_chosen_string(void *fdt, const char *name, const char *value) {
+    int chosen_offset = fdt_get_or_add_chosen_node(fdt);
+    if (chosen_offset < 0) {
+        return chosen_offset;
+    }
+
+    return fdt_setprop_string(fdt, chosen_offset, name, value);
+}
+
+int fdt_set_chosen_uint64(void *fdt, const char *name, uint64_t value) {
+    int chosen_offset = fdt_get_or_add_chosen_node(fdt);
+    if (chosen_offset < 0) {
+        return chosen_offset;
+    }
+
+    return fdt_setprop_u64(fdt, chosen_offset, name, value);
+}
+
+#endif
--- a/common/protos/linux_riscv64.c
+++ b/common/protos/linux_riscv64.c
@@ -1,0 +1,125 @@
+#if defined(__riscv64)
+
+#include <stdint.h>
+#include <stddef.h>
+#include <stdnoreturn.h>
+#include <protos/linux.h>
+#include <fs/file.h>
+#include <lib/libc.h>
+#include <lib/misc.h>
+#include <lib/term.h>
+#include <lib/config.h>
+#include <lib/print.h>
+#include <lib/uri.h>
+#include <mm/pmm.h>
+#include <sys/idt.h>
+#include <lib/fb.h>
+#include <lib/acpi.h>
+#include <lib/fdt.h>
+#include <libfdt/libfdt.h>
+
+// The following definitions and struct were copied and adapted from Linux
+// kernel headers released under GPL-2.0 WITH Linux-syscall-note
+// allowing their inclusion in non GPL compliant code.
+
+struct linux_header {
+    uint32_t code0;
+    uint32_t code1;
+    uint64_t text_offset;
+    uint64_t image_size;
+    uint64_t flags;
+    uint32_t version;
+    uint32_t res1;
+    uint64_t res2;
+    uint64_t res3;          // originally 'magic' field, deprecated
+    uint32_t magic2;
+    uint32_t res4;
+} __attribute__((packed));
+
+// End of Linux code
+
+#define LINUX_HEADER_MAGIC2             0x05435352
+#define LINUX_HEADER_MAJOR_VER(ver)     (((ver) >> 16) & 0xffff)
+#define LINUX_HEADER_MINOR_VER(ver)     (((ver) >> 0)  & 0xffff)
+
+noreturn void linux_load(char *config, char *cmdline) {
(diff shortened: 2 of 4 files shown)


common/lib/fdt.c
```diff
@@ -0,0 +1,35 @@
+#if !defined(__i386__) || !defined(__x86_64__)
+
+#include<stdint.h>
+#include<stddef.h>
+#include<libfdt/libfdt.h>
```
ElectrodeYT: Style issue?

ElectrodeYT (approved):
LGTM, also thanks for beating me to it :)

common/lib/fdt.c
ziyao233: got it. fixed fdt.h also.

common/lib/fdt.c
```diff
@@ -0,0 +1,35 @@
+#if !defined(__i386__) || !defined(__x86_64__)
```
Mintsuki: Shouldn't this be &&?

common/protos/linux_riscv64.c
```diff
@@ -0,0 +1,119 @@
...
+
+        ret = fdt_set_chosen_uint64(dtb, "linux,initrd-end", (uint64_t)(module_base + module_size));
+        if (ret < 0)
+            printv("linux: cannot set initrd parameter: %s\n", fdt_strerror(ret));
+
+    }
+
+    printv("linux: bsp hart %d, device tree blob at %x\n", bsp_hartid, dtb);
+
+    void (*kernel_entry)(uint64_t hartid, uint64_t dtb) = kernel_base;
```
Mintsuki, line 118: Should be marked as `noreturn`.

common/protos/linux_riscv64.c
```diff
@@ -0,0 +1,119 @@
...
+    asm volatile ("csrci   sstatus, 0x2\n\t"
+                  "csrw    sie, zero\n\t");
```
Mintsuki: Should have a `memory` clobber.

Mintsuki: Actually, if this is to be left as non-extended asm, then the volatile can just be dropped instead.

common/protos/linux_riscv64.c
ziyao233: A function pointer cannot be marked as noreturn, only definition and declaration could. A `__builtin_unreachable()` is inserted to prevent warnings.

common/protos/linux_riscv64.c
```diff
@@ -0,0 +1,119 @@
...
+    kernel_entry(bsp_hartid, (uint64_t)dtb);
+    __builtin_unreachable();
+}
+
+#endif
```
Mintsuki, line 125: This all looks good, but I'd prefer if it was added to `protos/linux.c` using `ifdef`s instead of into a new file.

common/protos/linux_riscv64.c
Mintsuki: That's odd, I've been able to use at least `__attribute__((noreturn))` on pointers before, unless it was just silently ignored.

Mintsuki: I am probably just misremebering, sorry.

Mintsuki:
In general it all LGTM, minus the comments I reviewed with. Amazing job!

I wonder if and how much of this would be portable to aarch64 as well...

common/lib/fdt.c
```diff
@@ -0,0 +1,35 @@
...
+    int offset = fdt_subnode_offset(fdt, 0, "chosen");
+
+    if (offset == -FDT_ERR_NOTFOUND) {
+        offset = fdt_add_subnode(fdt, 0, "chosen");
+    }
+
+    return offset;
+}
+
+int fdt_set_chosen_string(void *fdt, const char *name, const char *value)
```
Mintsuki: Style (braces)

common/lib/fdt.c
```diff
@@ -0,0 +1,35 @@
...
+{
+    int chosen_offset = fdt_get_or_add_chosen_node(fdt);
+    if (chosen_offset < 0)
+        return chosen_offset;
+
+    return fdt_setprop_string(fdt, chosen_offset, name, value);
+}
+
+int fdt_set_chosen_uint64(void *fdt, const char *name, uint64_t value)
```
Mintsuki: Style (braces)

common/protos/linux_riscv64.c
ziyao233: I think merging linux protocols related code on all architectures into a single file will result in a file contains `ifdefs` everywhere, which is less readable.
I would prefer renaming `protos/linux.c` as `protos/linux_x86.c` and extract arch-independent code into a `protos/linux.c`, which could be done in another PR.

common/protos/linux_riscv64.c
ziyao233: > Actually, if this is to be left as non-extended asm, then the volatile can just be dropped instead.

I wil do this.

ziyao233:
> In general it all LGTM, minus the comments I reviewed with. Amazing job!
> 
> I wonder if and how much of this would be portable to aarch64 as well...

Kernel image header on AArch64 is basically the same as RISCV's, but there is some different booting requirements.

I could create a new pull request later, extract some arch-independent code out then add aarch64 support as well, just as said before.

common/protos/linux_riscv64.c
Mintsuki: Alright, sure.

Mintsuki:
Sounds good! Linux booting on aarch64 has been something very desirable for a long while...

common/lib/fdt.c
```diff
@@ -0,0 +1,33 @@
...
+    if (offset == -FDT_ERR_NOTFOUND) {
+        offset = fdt_add_subnode(fdt, 0, "chosen");
+    }
+
+    return offset;
+}
+
+int fdt_set_chosen_string(void *fdt, const char *name, const char *value) {
+    int chosen_offset = fdt_get_or_add_chosen_node(fdt);
+    if (chosen_offset < 0)
```
Mintsuki: One last thing, style still, could you use braces even for single line ifs? Thank you! (This is done for all new code, and at some point I should go over the whole codebase to change it)

ziyao233: Okay. It is helpful to have explicit style documentation and style checker/formatter configuration (like `.clang-format`)

Mintsuki: I do not like it.

Mintsuki:
Thanks!

PR #372, opened 2024-06-16
Support booting on usb

firasuke:
This allows resulting ISO image files to run from usb thumb drives, which is a popular use case.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -66,6 +66,7 @@ Place any other file you want to be on the final ISO in said directory, then
 run:
 ```
 xorriso -as mkisofs -b <relative path of limine-bios-cd.bin> \
+        -hfsplus -apm-block-size 2048 \
         -no-emul-boot -boot-load-size 4 -boot-info-table \
         --efi-boot <relative path of limine-uefi-cd.bin> \
         -efi-boot-part --efi-boot-image --protective-msdos-label \


Mintsuki:
ISOs built as-is already support being booted off USB drives.

firasuke:
Yes but it depends on the initramfs implementation. This makes sure LABEL is being picked up correctly on hybrid boots.

Mintsuki:
Where did you obtain this flag from? What is the use case where you needed it, exactly?

firasuke:
> Where did you obtain this flag from? What is the use case where you needed it, exactly?

Attempting to produce an ISO without these flags will render the resulting drive where the ISO is being copied to (regardless of the tool `cat` or `dd`) not having a proper partition with a `LABEL` equivalent of the `volid` provided to the `xorriso` command.

e.g. (even after using a proper `volid` you can see the resulting partitions contain nothing that resembles the provided `LABEL` and thus can't be picked up by most initramfs implementations in live environments):

```sh
/dev/sda3: PARTLABEL="Gap1" PARTUUID="..."
/dev/sda2: SEC_TYPE="msdos" UUID="..." BLOCK_SIZE="512" TYPE="vfat" PARTLABEL="EFI boot partition" PARTUUID="..."
/dev/sda1: PARTLABEL="Gap0" PARTUUID="..."
```

Most initramfs generators rely on the `LABEL` of the partition to boot under live environments (a typical check for when `blkid -L LABEL` is available to be mounted or handled appropriately).

With these flags an additional partition is added:

```sh
/dev/sda4: BLOCK_SIZE="2048" LABEL="$VOLID" TYPE="hfsplus" PARTLABEL="HFSPLUS" PARTUUID="..."
```

Apparently this is also what `grub-mkrescue` does by default when it is run.

Edit 1: The resulting ISO works fine under virtualization without these flags as it is run with the proper `LABEL` and is dealt with as a typical `iso9660` filesystem, while on a USB thumb drive, that is not always the case as it might get picked up as `hfsplus`.

firasuke:
Glad this is finally fixed!

PR #387, opened 2024-07-20
docs: PROTOCOL.md: specify TCR_EL1/TTBR0_EL1 values for base revision > 0

Qix-:
As discussed in Discord, this further specifies the AArch64 entry state for the lower half of the memory map such that kernels targeting base revisions > 0 can more aptly create identity maps if need be.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -321,7 +321,10 @@ instructions or registers.
 
 Higher ELs do not interfere with accesses to the generic timer and counter.
 
-The used translation granule size is undefined.
+The used translation granule size for both `TTBR0_EL1` and `TTBR1_EL1` is 4KiB.
+
+For base revisions &gt; 0, `TCR_EL1.T0SZ` is set to 16, and `TTBR0_EL1`
+is left undefined.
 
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -304,6 +304,9 @@ its own.
 
 The `MAIR_EL1` register contents are described above, in the caching section.
 
+For base revisions &gt; 0, `TCR_EL1.tg0` is set to a 4KiB granule size,
+`TCR_EL1.t0sz` is set to 16, and `TTBR0_EL1` is left undefined.
```
qookei: A little lower the spec says:
> The used translation granule size is undefined.

IMO it'd be better to place this next to that. Also IMO tg0/t0sz should be uppercase (as it is in the ARMARM) but that's not a big deal.

Qix-: Is that better?

Qix-:
Alright after more deliberation in Discord it was decided to just go the whole way and specify 4KiB granule sizes in all cases. Latest changes should reflect that :)

qookei (approved):

PR #411, opened 2024-09-17
protos/multiboot2: fix issues with booting relocatable multiboot2 payloads

monkuous:
This PR fixes two issues that caused booting relocatable multiboot2 payloads to fail:
1. The order of the parameters to the ALIGN_* macros was incorrect (so alignment was aligned to min/max addr instead of the other way around). This caused incorrect alignment, or a division exception if lowest-to-highest allocation is preferred and the minimum address is zero.
2. `reloc_slide` was calculated as the absolute offset between the declared base address and the actual base address, but used as if it weren't absolute. This caused the relocated entry point address to be calculated incorrectly if highest-to-lowest allocation is preferred.

### DIFF
--- a/common/protos/multiboot2.c
+++ b/common/protos/multiboot2.c
@@ -314,24 +314,24 @@ noreturn void multiboot2_load(char *config, char* cmdline) {
         }
     }
 
-    uint64_t reloc_slide = 0;
+    int64_t reloc_slide = 0;
 
     if (has_reloc_header) {
         bool reloc_ascend;
-        uint64_t relocated_base;
+        int64_t relocated_base;
 
         switch (reloc_tag.preference) {
             default:
             case 0: case 1: // Prefer lowest to highest
                 reloc_ascend = true;
-                relocated_base = ALIGN_UP(reloc_tag.min_addr, reloc_tag.align);
+                relocated_base = (int64_t)ALIGN_UP(reloc_tag.min_addr, reloc_tag.align);
                 if (relocated_base + ranges->length > reloc_tag.max_addr) {
                     goto reloc_fail;
                 }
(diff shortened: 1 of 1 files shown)


Mintsuki:
Hi, thanks for the PR!

The first commit LGTM, but the second one would probably be best if changed to use a signed `int64_t` instead of a `uint64_t` for `reloc_slide` (and casting `relocated_base` to `int64_t` as well in the assignment).

monkuous:
Thanks for the feedback. I opted not to use `int64_t` in this case because both its calculation and usage relies on integer overflow, which is undefined behavior for signed integers.

Mintsuki:
I'm not sure what's the signed overflow? From a quick look it only seems to cross the negative<->positive barrier which is perfectly defined behaviour for signed integers (and not considered overflow). I may have missed something though.

monkuous:
Nevermind, now that I'm looking at it again, you're right. I'll change it to be signed.

common/protos/multiboot2.c
```diff
@@ -314,24 +314,24 @@ noreturn void multiboot2_load(char *config, char* cmdline) {
         }
     }
 
-    uint64_t reloc_slide = 0;
+    int64_t reloc_slide = 0;
 
     if (has_reloc_header) {
         bool reloc_ascend;
-        uint64_t relocated_base;
+        int64_t relocated_base;
```
Mintsuki: I wouldn't change this to signed. As I said previously I'd cast it to `int64_t` on line 359.
Aka `reloc_slide = (int64_t)relocated_base - ranges->target;`

Mintsuki:
LGTM thanks!

PR #413, opened 2024-09-28
protos/limine: increase MEMMAP_MAX

monkuous:
On some systems, the memory map exceeds 256 entries (on my system, it has 312 entries). That causes the bootloader to panic when booting Limine payloads. This PR simply doubles the limit.

### DIFF
--- a/common/protos/limine.c
+++ b/common/protos/limine.c
@@ -34,7 +34,7 @@
 
 #define MAX_REQUESTS 128
 
-#define MEMMAP_MAX 256
+#define MEMMAP_MAX 512
 
 static int paging_mode;
 
--- a/common/protos/limine.c
+++ b/common/protos/limine.c
@@ -34,7 +34,7 @@
 
 #define MAX_REQUESTS 128
 
-#define MEMMAP_MAX 512
+#define MEMMAP_MAX 1024
 
 static int paging_mode;
 


Mintsuki:
Hello, thanks for the PR!

May I know what system is this, and whether we're talking BIOS or UEFI? Thanks.

monkuous:
- CPU: AMD Ryzen 7 5800X
- GPU: AMD Radeon RX 6500 XT
- Motherboard: Gigabyte B550M S2H (firmware version F18)
- RAM: 64 GiB

The issue only occurs in UEFI mode. The memory map mostly consists of small `LIMINE_MEMMAP_USABLE` entries, each followed by a small `LIMINE_MEMMAP_BOOTLOADER_RECLAIMABLE` entry.

Mintsuki:
Is there a chance you could post the full memory map alongside mentioning what kernel specifically you were trying to load and if it does anything unusual (like loading lots of modules, etc)?

monkuous:
I was attempting to load a custom closed-source kernel. It doesn't do anything special, and I managed to reproduce the issue by adding a memory map request to `limine-c-template-x86-64`. Here's the Limine config I used, as well as the memory map dump: https://hst.sh/onirisinog.yaml. Note that the memory map was transcribed using OCR, so it might have small errors, though I haven't caught any during a quick look though the transcribed text.

Mintsuki:
Thanks. May I also know the Limine version used just in case?

monkuous:
I initially encountered it on 8.0.13, but while diagnosing the issue I was using https://github.com/limine-bootloader/limine/commit/7d66858bcb75995625360c462eff925bd45f9a9e

Mintsuki:
Do you think it would be possible for you to join the Limine Discord server linked in the README.md file in order to run some live tests on that machine?

I have no issues with this PR, but my main concern is that this may be some band aid fix for another underlying Limine issue (triggered by that specific machine's firmware/memory map). So in that case, I'd rather fix the underlying issue.

PR #439, opened 2024-12-20
protos/linux_risc: add LoongArch support

ziyao233:
This series cleans up linux_risc.c, implements EFI-table-based boot param passing and finally supports Linux boot protocol on LoongArch.

Has been tested with EDK2 firmware on qemu-emulated loongarch64 and riscv64, everything works. I'll verify on loongarch64/aarch64 real hardware when I'm available.

Thanks for your time and review.

### DIFF
--- a/common/lib/misc.c
+++ b/common/lib/misc.c
@@ -16,7 +16,7 @@ EFI_BOOT_SERVICES *gBS;
 EFI_RUNTIME_SERVICES *gRT;
 EFI_HANDLE efi_image_handle;
 EFI_MEMORY_DESCRIPTOR *efi_mmap = NULL;
-UINTN efi_mmap_size = 0, efi_desc_size = 0;
+UINTN efi_mmap_size = 0, efi_desc_size = 0, efi_mmap_key = 0;
 UINT32 efi_desc_ver = 0;
 #endif
 
@@ -205,9 +205,8 @@ bool efi_exit_boot_services(void) {
 
     EFI_MEMORY_DESCRIPTOR tmp_mmap[1];
     efi_mmap_size = sizeof(tmp_mmap);
-    UINTN mmap_key = 0;
 
-    gBS->GetMemoryMap(&efi_mmap_size, tmp_mmap, &mmap_key, &efi_desc_size, &efi_desc_ver);
+    gBS->GetMemoryMap(&efi_mmap_size, tmp_mmap, &efi_mmap_key, &efi_desc_size, &efi_desc_ver);
 
     efi_mmap_size += 4096;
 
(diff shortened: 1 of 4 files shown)


qookei:
There is a regression with booting Linux on AArch64 with this PR. I haven't investigated what exactly goes wrong, but Linux panics(?) and goes into an infinite loop before it even prints anything to `earlycon` (but after it has set up paging and put itself in the higher half etc).

ziyao233:
> There is a regression with booting Linux on AArch64 with this PR. I haven't investigated what exactly goes wrong, but Linux panics(?) and goes into an infinite loop before it even prints anything to `earlycon` (but after it has set up paging and put itself in the higher half etc).

Thanks for the feedback.

I have tried this kernel[1] with EDK2 in qemu-system-aarch64, it boots when ACPI is set to off (with `-machine virt,acpi=off`). The ACPI-enabled case doesn't work even in the trunk.

What kernel are you trying to boot? A binary file together with the `.config` will be helpful, thanks.

[1]: https://os-repo-cn.ewe.moe/eweos/main/os/aarch64/linux-6.12.5-1-aarch64.pkg.tar.zst

qookei:
> What kernel are you trying to boot?

I'm trying kernel 6.9.7-1-aarch64-ARCH from Arch Linux ARM (just what I had lying around from when I was working on the AArch64 Linux boot protocol support). With the following config it gets far enough to display the 4 penguins on the framebuffer, and panic because it can't mount the root partition.
```
/Linux Test
    protocol: linux
    kernel_path: boot():/boot/Image
    cmdline: earlycon=pl011,mmio32,0x9000000
```

> The ACPI-enabled case doesn't work even in the trunk.

The kernel I'm testing with boots fine without adding `,acpi=off`, and is actually using ACPI according to the kernel logs.

qookei:
The kernel you have provided also works with ACPI enabled, are you perhaps using an older/newer version of QEMU? I'm testing with QEMU 8.2.3. It also doesn't boot with this PR applied for me.

ziyao233:
> > What kernel are you trying to boot?
> 
> I'm trying kernel 6.9.7-1-aarch64-ARCH from Arch Linux ARM (just what I had lying around from when I was working on the AArch64 Linux boot protocol support). With the following config it gets far enough to display the 4 penguins on the framebuffer, and panic because it can't mount the root partition.
> 
> ```
> /Linux Test
>     protocol: linux
>     kernel_path: boot():/boot/Image
>     cmdline: earlycon=pl011,mmio32,0x9000000
> ```
> 
> > The ACPI-enabled case doesn't work even in the trunk.
> 
> The kernel I'm testing with boots fine without adding `,acpi=off`, and is actually using ACPI according to the kernel logs.

My bad, trunk should work. I might mess something up when doing a bisect.

Please try the new HEAD and see whether it's fixed now (it should be).

For the details: currently aarch64 relies a working dtb to pass memory mapping and EFI table. If limine doesn't get one, it was still possible to create one from scratch and fill EFI tables and memory mappings into it. My new code breaks the creation.

qookei (approved):
Yep, works now, thanks for the PR! And yeah, AArch64 needs at least a dummy DTB with just the cmdline and pointers to the UEFI tables (this is what Linux's EFISTUB does internally as well). IIRC so does RISC-V, but I guess the default is no ACPI there.

ziyao233:
> For the details: currently aarch64 relies a working dtb to pass memory mapping and EFI table. If limine doesn't get one, it was still possible to create one from scratch and fill EFI tables and memory mappings into it. My new code breaks the creation.

Now we always have a working dtb (got from VendorTable or created from scratch), so some extra checks could be removed. A new HEAD is pushed.

qookei:
We could probably factor out loading the DTB from file and getting it via `get_device_tree_blob` into a helper function to avoid allocating memory for the DTB if we end up using the one from the file (like what `prepare_device_tree_blob` on trunk does at the start).

ziyao233:
FYI, I have got linux kernel booted on loongarch64 real machine with limine 8.6.1 and this patch[1] reverted. I guess it's okay to enable loongarch64 linux boot protocol by default now :)




[1]: https://github.com/limine-bootloader/limine/commit/c73505f13b5128470c16f370cfd2a6d4425cfa8d

PR #442, opened 2024-12-28
lib: Add support for loading PE images

48cf:


common/lib/pe.c
```diff
@@ -0,0 +1,383 @@
...
+        *_ranges_count = range_count;
+
+        for (size_t i = 0, j = 0; i < nt_hdrs->FileHeader.NumberOfSections; i++) {
+            IMAGE_SECTION_HEADER *section = &sections[i];
+
+            if (section->Characteristics & IMAGE_SCN_MEM_DISCARDABLE) {
+                continue;
+            }
+
+            ranges[j].base = *virtual_base + ALIGN_UP(section->VirtualAddress, alignment);
```
Qwinci: shouldn't this be ALIGN_DOWN?

common/lib/pe.c
```diff
@@ -0,0 +1,383 @@
...
+            ranges[j].length = ALIGN_UP(section->VirtualSize, alignment);
```
Qwinci: might want to add the misalign from the base here before aligning up too.

common/lib/pe.c
```diff
@@ -0,0 +1,383 @@
...
+
+        if (import_desc->Name != 0) {
+            panic(true, "pe: Kernel must not have any imports");
+        }
+    }
+
+    if (reloc_dir->VirtualAddress != 0) {
+        IMAGE_BASE_RELOCATION_BLOCK *block = (IMAGE_BASE_RELOCATION_BLOCK *)(*physical_base + reloc_dir->VirtualAddress);
+
+        while (block->VirtualAddress != 0) {
```
Qwinci: don't think this is guaranteed, you should instead use reloc_dir->size and check the offset you are into the section.

Qwinci (changes requested):

PR #493, opened 2025-04-12
libc: add a strcat implementation

braewoods:
I was working on a new PR but found there was no strcat for string building. This will resolve that.


braewoods:
I rebased on trunk so no idea why it is still saying that.

PR #496, opened 2025-04-13
Read config file snippets from limine.d

marv7000:
This PR adds support for split config files located in a `limine.d` directory.

TODOs:
- [x] readdir for FAT
- [ ] readdir for ISO9960
- [ ] readdir for TFTP

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,9 @@
 *.dtb
 *~
 
+# Keep limine.d, since the .d conflicts with the '*.d' rule
+!test/limine.d
+
 /bin
 /build
 /toolchain-files
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -25,6 +25,13 @@ candidates in subsequent partitions or directories are ignored.
 It is thus imperative that the intended config file is placed in a location
 that will not be shadowed by another candidate config file.
 
+## Split config files
+
+Limine allows loading multiple additional config file snippets from a directory
+named `limine.d/`. All files in that directory will be appended to
+`limine.conf`, sorted by their file name in ascending order.
+`limine.d/` has to be located in the same directory as `limine.conf`.
+
 ## Structure of the config file
 
 The Limine configuration file is comprised of *menu entries* and *options*.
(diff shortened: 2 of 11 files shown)


lzcunt:
How is config file hash validation going to work with this?

How will entry ordering work, which is also important for setting the default entry? Will global keys set by drop-ins be overwritten by drop-ins with a higher priority? Will drop-ins be able to set global keys at all?

IMO adding a new drop-in should not cause the default entry to change due to all the indexes moving around. The entry index could be scoped to the file to prevent that, but that won't work with a simple concatenation of all files. Maybe the entry index for the default entry could be replaced by a unique entry key/id instead (defaulting to the entry name, but could be overwritten with an entry option)

BTW thanks for this PR, this is something I'm looking forward to

marv7000:
The easiest way would be to just have limine.conf hold a hash that also contains all limine.d entries. That way you can also validate that there aren't any unwanted changes to any of the files.

Drop-ins can set global keys without any problem. That way you could have a file that contains only variable declarations.

Maybe the default entry selection could be done by name rather than by index. But I'm not too sure about that.

lzcunt:
> The easiest way would be to just have limine.conf hold a hash that also contains all limine.d entries. That way you can also validate that there aren't any unwanted changes to any of the files.

We could implement that logic in the host tool, rather than polluting limine.conf. As in, the host tool could hash all the drop-ins and the limine.conf file together, then bake *that* into the EFI executable.

> Drop-ins can set global keys without any problem. That way you could have a file that contains only variable declarations.

Cool

> Maybe the default entry selection could be done by name rather than by index. But I'm not too sure about that.

By name could cause issues, e.g. sub-entries could reasonably have the same name but different parents. So we should have unique identifiers/keys for entries if we want to do this. I think systemd-boot does it this way too? But they also have a lot more complicated sorting mechanisms.

marv7000:
> As in, the host tool could hash all the drop-ins and the limine.conf file together, then bake that into the EFI executable.

Yes, that was what I meant.

qookei:
Development has moved to Codeberg: https://codeberg.org/Limine/Limine. Please reopen this pull request there.

Issue #528, opened 2025-08-15
Limine 9.6.0 fails to boot Shim


JakubVanek:
Hello,

I am trying to set up Limine on my system in a somewhat atypical configuration:

<img width="307" height="391" alt="Image" src="https://github.com/user-attachments/assets/8cf5524d-d8f6-4129-beeb-29a4b4a9990f" />

My goals with this boot path are the following:
- I would like to keep Secure Boot enabled and working
  - Once I finish configuring Limine I will enroll the configuration hash into the binary and sign that
  - I also would like UEFI to check signatures on EFI binaries that Limine loads
- I would like to preserve as much as possible from the distribution boot flow (so that I do not have to re-sign GRUB or kernels on APT updates)
- I would like to avoid booting Windows through GRUB to avoid the Bitlocker recovery screen when Ubuntu decides to update GRUB.
  - Yet I would like to preserve the Bitlocker TPM PCR[4] binding

I was able to make this work with systemd-boot without much effort, but with Limine 9.6.0 I am facing problems. I am using the following config:
```
timeout: 5
term_font_scale: 2x2

/Linux
    protocol: efi
    path: boot(1):/EFI/ubuntu/shimx64.efi

/Windows
    protocol: efi
    path: boot(1):/EFI/Microsoft/Boot/bootmgfw.efi
```

When I try to load `shimx64.efi` via Limine, the boot fails because `shimx64.efi` fails to find `grubx64.efi` (the screenshot is from QEMU, but a similar error appears when I try this directly on my laptop):

<img width="655" height="202" alt="Image" src="https://github.com/user-attachments/assets/a7d1639f-bf07-4774-8455-af4333c2ed70" />

Yet, if I boot `shimx64.efi` via systemd-boot or via the UEFI shell, it starts successfully and boots into Linux.

The issue can be reproduced using the following disk image: [disk.img.gz](https://github.com/user-attachments/files/21792386/disk.img.gz). If you have QEMU and OVMF installed, it can be booted by running
```
qemu-system-x86_64 -bios /usr/share/ovmf/OVMF.fd -hda disk.img
```

I have bisected the problem and the first bad commit is https://github.com/limine-bootloader/limine/commit/030aa9ec91ef56661c8c7a6805abbedfec858af6 - Limine before this commit can boot Shim successfully. Also, systemd-boot (which works in my setup) handles loading binaries similarly to Limine before that commit - it loads the binary directly from the filesystem via `LoadImage()`.

[...]

PR #529, opened 2025-08-16
protos/efi: Make the file path passed to executed binaries relative

JakubVanek:
This PR fixes Limine chainloading of Shim. The issue was that Limine was setting the `new_handle_loaded_image->FilePath` to a path that included the device path of the partition (meaning the full path was something like `PciRoot(0)\Pci(0x1,0x1)\Ata(Primary,Master)\HD(1,GPT,4B0E42EC-AC3C-4E2E-B700-1EE0ECA3BEC3)\EFI\ubuntu\shimx64.efi`). However, it seems that the path should instead be relative to the `new_handle_loaded_image->DeviceHandle`. The [EFI online docs](https://uefi.org/specs/UEFI/2.10/09_Protocols_EFI_Loaded_Image.html) say
> FilePath ... A pointer to the file path portion specific to DeviceHandle that the EFI Image was loaded from.

This PR modifies the code to instead set the path to something like `EFI\ubuntu\shimx64.efi`. With this change, Shim can successsfully locate itself and load the GRUB binary from the directory where it is placed.

Fixes https://github.com/limine-bootloader/limine/issues/528

### DIFF
--- a/common/protos/chainload.c
+++ b/common/protos/chainload.c
@@ -195,71 +195,56 @@ load:
 
 #elif defined (UEFI)
 
-/*
-static void devpath_print(EFI_DEVICE_PATH_PROTOCOL *DevicePath) {
-    for (;;) {
-        uint8_t type = DevicePath->Type;
-        uint8_t subtype = DevicePath->SubType;
-        size_t length = *(uint16_t *)DevicePath->Length;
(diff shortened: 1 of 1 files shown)


Mintsuki:
LGTM

PR #533, opened 2026-02-02
riscv: implement idempotent-op based cache flushing for non-Zicbom SoCs

FengHeting:
On RISC-V 64 platforms lacking the Zicbom extension (such as the StarFive JH7110), the standard cbo.flush instruction cannot be used for cache management, which can lead to data inconsistency between the L2 cache and DDR in framebuffer scenarios.

This patch implements a rollback flush scheme based on equivalent read-modify-write (p[i] ^= 0), performing an in-place RMW operation on each 64-byte cache line to ensure that the cache line is marked as dirty and written back to memory during subsequent eviction. Combined with fence rw, rw, this method has been tested on VisionFive 2 and effectively eliminates visible display remnants.

### DIFF
--- a/common/lib/gterm.c
+++ b/common/lib/gterm.c
@@ -477,6 +477,15 @@ static void riscv_flush_callback(volatile void *base, size_t length) {
         asm volatile("cbo.flush (%0)" :: "r"(ptr) : "memory");
     }
 }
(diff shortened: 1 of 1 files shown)


marv7000:
LGTM but please open this PR upstream at https://codeberg.org/Limine/Limine and make it target the branch `trunk`. Thank you!

FengHeting:
Okay

PR #543, opened 2026-04-10
decompressor: gzip/tinf -> limlz

iczelia:
removes external dependency on tinf by replacing the compression algorithm with a simpler, faster, smaller and more auditable fixed-width LZ77 encoding purpose-tailored to x86 code mixed with data.

before: decompressor.bin 2,492 bytes (tinf dependency) with .text 0x875 and .rodata 0x13c bytes each.
after: decompressor.bin consists only of .text, 0xe6-byte decompressor; 90.8% reduction in decompressor volume.

the dependency on gzip during compile-time is replaced by host/limlzpack.c, a Lempel-Ziv encoder in 275 SLoC that uses a suffix array matchfinder (prefix-doubling in mathcal O(n log^2 n) and Storer-Szymanski backwards parse. the fixed-width formats packets as [F][LLLL][MMM], favouring a literal-skewed distribution with F switching between one-byte and two-byte offsets (favouring recent statistics).

integrity checking is done via crc32 with the polynomial 0xEDB88320, reflected.

the effective loss in compression ratio by using a tremendously simpler and less packed with edge cases algorithm causes a compression ratio hit well below 1KB, factoring in the stub sizes.

this algorithm, largely, does not generalize - in terms of compression ratio - to other inputs than Limine 2nd stage code (x86 code + data); it has been largely tailored for this specific purpose only.


Mintsuki:
LGTM, thank you very much!

Ari4ka (approved):

PR #544, opened 2026-04-11
crypto: streaming blake2b for validation

iczelia:
exposes the internals of the streaming blake2b checksum and integrates the cryptography code with the standard fake-vtable mechanism for file i/o in Limine, allowing for validation of the checksum without reading the whole file into memory in one shot and caching it until closed.


Mintsuki:
LGTM, thanks!

Issue #541, opened 2026-04-08
Feature Request: Add mechanism to show menu entries conditionally


BrainwreckedTech:
Currently, Limine will hide `protocol: bios` entries when booted from UEFI, and `protocol: efi` entries when booted from BIOS. This proves insufficient for Memtest86+, which uses `protocol: linux` for BIOS machines and `protocol: efi` for UEFI machines. This means the Memtest86+ entry for BIOS machines is always shown. Limine has the `${ARCH}` and `${FW_TYPE}` macros, but there's no way to act on them.

I suggest adding showonly: and maybe dontshow: which would take the values from `${ARCH}` and `${FW_TYPE}` and act accordingly.

Example:

```
/Memtest86+
    showonly: BIOS
    # or dontshow: UEFI
    protocol: linux
    path: boot():/memtest86+/memtest.bin
```
NOTE: the UEFI equivalent already hides itself on BIOS boots.

Example for having 32-bit and 64-bit OSes on the same drive:

```
/+Arch Linux x64v1
showonly: x86-64
comment: All AMD and Intel 64-bit CPUs
comment: machine-id: <uuid>

...

/+Arch Linux i786
showonly: x86-64 ia-32
comment: Intel Pentium 4, Pentium M, and Atom CPUs
comment: machine-id: <uuid>

...
```

In the first example, the wanted behavior is to hide the BIOS version of Memtest86+ on UEFI systems because you should use the UEFI version instead. But it won't be hidden becuase `protocol: linux` shows up on both BIOS and UEFI. Meanwhile, the UEFI equivalent gets hidden on BIOS systems because that uses `protocol: efi`, which is the wanted behavior as a BIOS system wouldn't even know what to do with an EFI binary.

In the second example, the showonly is used to prevent the option for the x86-64 OS from showing up on ia-32 CPUs. The annoyance here is that going ahead and trying to launch an x86-64 OS on IA-32 results in a silent and permanent hang, which is indistinguishable from older hardware just taking its sweet time to load the initramfs because, well, it's old hardware.

Issue #537, opened 2026-04-07
Support limiting boot menu entries to a specific boot mode and UEFI architecture


nl6720:
Imagine that you have tree binaries: `memtest.bin` (for BIOS booting on x86_64) `memtest86x64.efi` (for x64 UEFI), `memtest86ia32.efi` (for IA32 UEFI). If you add menu entries for all of them, then all will be shown in the boot menu. This clutters the boot menu since only one of them works.

It would be useful to have an option that specifies the boot mode and UEFI architecture for a menu entry.

----

Originally posted to https://codeberg.org/Limine/Limine/issues/11

PR #560, opened 2026-04-26
add options that allow hiding entries depending of firmware/architecture

PatrickHechler:
the new option `if_fw_type` can be set to either `bios` or `efi`/`uefi`.

+ `bios`: hide the entry, unless booted from bios
+ `efi`/`uefi`: hide the entry, unless booted from uefi

the new option `if_arch` is interpreted as a space seperated list of permitted architectures.    
the entry will be hidden if `${ARCH}` is not inside the list.    
note that for 32-bit executables the 64-bit architecutre and the 32-bit architecture needs to be in the list (i.e. `ia-32 x86-64`)

this is requested by #541 and #537


iczelia (changes requested):
Thanks for the patch. Unfortunately, by my understanding, this is completely wrong. Three main issues:

 - The `IF_FW_TYPE` block in `should_skip_entry` compares `cur_entry_protocol` instead of `cur_entry_if_fw_type`; the new option is never actually read, and entries without a `PROTOCOL` key will crash on the dereference of a `NULL` pointer.
 - I don't understand the purpose or even the origin of the wait macro and the val declaration above it.
 - You have not documented your changes in CONFIG.md. The architecture detection block also seems to be a verbatim copy-paste of common/lib/config.c:415-434, perhaps a shared helper is in order.

PatrickHechler:
thanks for the quick response
> Thanks for the patch. Unfortunately, by understanding, this is completely wrong. Three main issues:
> 
>     * The `IF_FW_TYPE` block in `should_skip_entry` compares `cur_entry_protocol` instead of `cur_entry_if_fw_type`; the new option is never actually read, and entries without a `PROTOCOL` key will crash on the dereference of a `NULL` pointer.
sorry for that, should be fixed now
> 
>     * I don't understand the purpose or even the origin of the wait macro and the val declaration above it.
sorry for that, the wait macro was just a breakpoint for debugging
> 
>     * You have not documented your changes in CONFIG.md. The architecture detection block also seems to be a verbatim copy-paste of common/lib/config.c:415-434, perhaps a shared helper is in order.
I have added documentation to CONFIG.md now.    
I didn't want to modify some other place, but yes, it is mostly copy-pasted from config.c and a `#define ARCH` somewhere may be better (where?).

CONFIG.md
```diff
@@ -207,6 +207,11 @@ Editor control options:
 * `cmdline` - The command line string to be passed to the kernel/executable.
   Can be omitted.
 * `kernel_cmdline` - Alias of `cmdline`.
+* `if_fw_type` - Hide the entry if the firmware does not match. Valid values
+  are: `bios`, `efi` (or `uefi`)
+* `if_arch` - Hide the entry if the current CPU architecture is not in the
+  space seperated list of permitted architecutres
```
iczelia: seperated -> separated.

iczelia: architecutres -> architectures

common/menu.c
```diff
@@ -617,6 +619,70 @@ static inline bool should_skip_entry(struct menu_entry *entry) {
             return true;
         }
     }
+    char *cur_entry_if_fw_type = config_get_value(entry->body, 0, "IF_FW_TYPE");
+    if (cur_entry_if_fw_type) {
+#if defined (UEFI)
+        if (strcmp(cur_entry_if_fw_type, "efi") != 0
+         && strcmp(cur_entry_if_fw_type, "uefi") != 0) {
```
iczelia: The new option accepts bios/efi/uefi but the existing `FW_TYPE` produces BIOS/UEFI. Please either accept both (case-insensitive compare) or switch to the canonical spellings.

common/menu.c
```diff
@@ -617,6 +619,70 @@ static inline bool should_skip_entry(struct menu_entry *entry) {
...
+        if (strcmp(cur_entry_if_fw_type, "bios") != 0) {
+#else
+#error "Unspecified firmware type"
+#endif
+            return true;
+        }
+    }
+    char *cur_entry_if_arch = config_get_value(entry->body, 0, "IF_ARCH");
+    if (cur_entry_if_arch) {
+        const char *arch;
```
iczelia: the deduplicated function can likely take the prototype of `static inline const char *current_arch(void)` in `common/sys/cpu.h`.

iczelia (changes requested):
There are other outstanding issues.

common/menu.c
PatrickHechler: using the same value as `FW_TYPE` makes sense, I just used the bios/efi protocol values (except for the _chainloading values).    
should this then too be extracted into a function? (for now I have placed the function `current_firmware` in `misc.h`)

CONFIG.md
PatrickHechler: should be fixed now

common/menu.c
PatrickHechler: should be now now

iczelia (approved):
OK, this is good enough now.

PatrickHechler:
thanks for the help

Issue #576, opened 2026-05-28
Support for non-rebooting UEFI handover to prevent PCR alterations


valiukasd:
Hello! I am a developer using my machine for programming, and I recently decided to dual-boot Windows 11 and CachyOS. I wanted a modern, fast setup, so after some digging I chose Limine alongside CachyOS's automated Snapper configuration. I am using Secure Boot with custom certificates signed via sbctl, and my LUKS keys are bound to my TPM.

While CachyOS boots flawlessly, I ran into a major roadblock with Windows: every time I boot Windows through Limine, BitLocker forces me into the recovery screen to enter my long recovery key.

I don't know a whole lot about the deep inner workings of bootloaders, but after a lot of troubleshooting and researching, I discovered why this happens. Limine's current protocol: efi reads the Windows boot manager into memory and handles the execution chain internally. Because of this, the TPM logs Limine's footprint and alters PCR 4 (the bootloader path). Windows sees this alteration, assumes the boot chain was tampered with, and locks me out.

I know Limine features the protocol: `efi_boot_entry` option, which uses `BootNext` to force a warm restart into Windows. While this successfully keeps the TPM path clean and stops BitLocker from tripping, it forces my motherboard to POST twice. This adds an extra 5 to 10 seconds to the boot time every time I want to switch to Windows to work or test code.

On the other hand, a bootloader like **rEFInd** avoids this by cleanly unloading itself and handing execution completely to the native UEFI image loader without forcing a hardware reset.

Proposed feature
To keep Limine's codebase clean and minimalist while giving dual-booters a smooth experience, I would love to propose adding a third EFI protocol option: protocol: `efi_handover` (or a similar flag/protocol implementation).

Instead of Limine parsing the target binary file and allocating memory itself, this protocol would defer directly to the motherboard's built-in UEFI image framework.

Conceptual Implementation:
When protocol: `efi_handover` is specified with a path, Limine would leverage standard UEFI firmware functions:

Translate the configuration text path into a native UEFI device path.

Call `gBS->LoadImage()` so the system firmware natively registers and caches the target binary (e.g., Windows Boot Manager).

Call `gBS->StartImage()` to cleanly terminate Limine's execution loop and hand off the boot phase.

[...]

PR #578, opened 2026-05-29
efi: add protocol: efi_handover for firmware-native chainloading

alexvoste:
### Description

   This PR implements the new `protocol: efi_handover` option proposed in the active BitLocker/TPM issue.

   ### Why this is needed:
   Currently, the standard `protocol: efi` reads the target binary into a memory buffer before passing it to `gBS->LoadImage()`. This execution from an anonymous RAM buffer alters the TPM PCR 4 measurement because Limine leaves its active footprint in the boot chain. As a result, Windows BitLocker detects a "tampered" boot path and forces the user into the recovery key screen on every dual-boot start.

   ### What this PR does:
   Instead of allocating memory and parsing the image manually, `efi_handover` uses native UEFI firmware services:
   1. It resolves the relative `EFI_DEVICE_PATH` (`efi_file_path`) and partition handle of the target image.
   2. It closes the file handle in Limine immediately, keeping the RAM clean.
   3. It passes the raw `EFI_DEVICE_PATH` directly to `gBS->LoadImage(FALSE, ...)`, allowing the motherboard's native filesystem drivers to handle the load and signature checks natively.
   4. It starts the image cleanly via `gBS->StartImage()`.

   This keeps the TPM measurements clean, allowing seamless dual-booting with Windows BitLocker active, without requiring double-POST reboots via `efi_boot_entry`.

   ### Testing status:
   - The code builds flawlessly under `clang -target x86_64-unknown-none-elf`.
   - I am running a pure Arch Linux system on my bare-metal machine without Windows, so I was unable to personally test the BitLocker bypass.
   - *I am opening this as a Draft PR* so the author of the issue or other contributors with dual-boot Windows + TPM configurations can test the compiled binary and confirm the fix.

### DIFF
--- a/common/menu.c
+++ b/common/menu.c
@@ -770,7 +770,8 @@ static inline bool should_skip_entry(struct menu_entry *entry) {
         if (strcmp(cur_entry_protocol, "efi") == 0
          || strcmp(cur_entry_protocol, "uefi") == 0
          || strcmp(cur_entry_protocol, "efi_chainload") == 0
-         || strcmp(cur_entry_protocol, "efi_boot_entry") == 0) {
+         || strcmp(cur_entry_protocol, "efi_boot_entry") == 0
+         || strcmp(cur_entry_protocol, "efi_handover") == 0)	 {
 #endif
             return true;
         }
@@ -2072,6 +2073,9 @@ noreturn void boot(char *config) {
 #if defined (UEFI)
     else if (!strcmp(proto, "efi_boot_entry")) {
         efi_boot_entry(config);
+    }
+    else if (!strcmp(proto, "efi_handover")) {
+	    chainload_handover(config, cmdline);
     }
 #endif
 
--- a/common/protos/chainload.c
+++ b/common/protos/chainload.c
@@ -256,6 +256,99 @@ static EFI_DEVICE_PATH_PROTOCOL *build_relative_efi_file_path(struct file_handle
     return device_path;
 }
 
+
+noreturn void chainload_handover(char *config, char *cmdline) {
+    // 1. Retrieve the target image path from config
+    char *image_path = config_get_value(config, 0, "PATH");
+    if (image_path == NULL) {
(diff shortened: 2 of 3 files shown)


Mintsuki:
Hey, thanks for the PR!

First off let me mention issue #576 so the author can get notified about this PR.

Second, I would frankly prefer to avoid adding yet another completely sepearate `efi_*` protocol. I already thought `efi_boot_entry` might have been a bit much, but this seems completely redundant to add.

If anything, I'd change the `efi` protocol itself to use this method rather than adding a completely new protocol. I'd be okay with that after careful consideration about potential breakages that might happen by doing this (something that comes to mind is that the `$`-prefixed transparent decompression would no longer work, for the few-to-no users actually using that with the `efi` protocol).

If needed it could also be made into a flag/modifier option inside the `efi` protocol itself, to preserve backwards compatibility while still avoiding a lot of code duplication.

In any case @valiukasd should confirm that it actually does achieve what we intend it to, before moving on.

alexvoste:
Hey! Thanks for the quick feedback, Mintsuki.

 I completely agree with keeping the protocol list clean and avoiding config bloating. Integrating this behavior directly into the existing `efi` protocol makes total sense.

 Personally, I think Option 2 (adding a flag/modifier option inside the existing `efi` protocol) is the safest bet. It keeps the codebase clean, avoids duplication, and preserves backwards compatibility (including that transparent `$`-decompression feature for the few users who rely on it).

 If we go with a flag, what name would you prefer? Something like `efi_handover: yes` in the config, or maybe a path modifier?

 I'll refactor the code as soon as we agree on the design. Let's also wait for @valiukasd to confirm the BitLocker/TPM bypass on their machine!

valiukasd:
Hey, I will test the feature as soon as I figure out how to build and use the local version of the bootloader. As for the flag or protocol, from a user standpoint, a flag seems more logical since the protocol is still the same. Will report here after I confirm if the changes solve the issue

PR #581, opened 2026-06-04
tools/limlzpack: simplify endian logic

sertonix:
I feel like using byteswap methods commonly leads to more complicated code than necessary.


iczelia:
This change is, of course, correct; the original endianness patch was not done by me despite me being the author of the code. It's a very roundabout and not very good way to do this so of course the suggested replacement is better. Please open a new pull request against trunk.

PR #588, opened 2026-06-11
protos/chainload: Add native_load option for efi protocol

### LINKED ISSUE #576: Support for non-rebooting UEFI handover to prevent PCR alterations

valiukasd:
Fixes and closes #576 

Implements a small code change to load the image via native drivers, if a flag **"native_load"** is provided. Documentation is updated, no breaking changes for existing users

Testing status:
- [x] Project builds
- [x] `protocol: efi` works without the flag **"native_load"**
- [x] `protocol: efi` works with the flag **"native_load"**
- [ ] `protocol: efi` works with the flag **"native_load"** and with hashing (#)
- [ ] `protocol: efi` works with the flag **"native_load"** and with transparent (de)compression ($)

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -287,6 +287,9 @@ Editor control options:
   * `resolution` - The resolution to be used. This setting takes the form of
     `<width>x<height>x<bpp>`. If the resolution is not available, Limine will
     pick another one automatically. Omitting `<bpp>` will default to 32.
+  * `native_load` - If set to `yes`, chainloads the EFI application natively
+    without using an intermediate memory buffer. This ensures TPM PCR 4
+    measurements reflect the exact on-disk binary (e.g. for BitLocker).
 
 * EFI Boot Entry protocol:
   * `entry` - The name of the EFI boot entry to reboot into.
--- a/common/protos/chainload.c
+++ b/common/protos/chainload.c
@@ -204,6 +204,44 @@ load:
 
 #elif defined (UEFI)
 
+static EFI_DEVICE_PATH_PROTOCOL *append_device_path(EFI_DEVICE_PATH_PROTOCOL *dp1, EFI_DEVICE_PATH_PROTOCOL *dp2) {
+    if (dp1 == NULL) {
+        return dp2;
+    }
+    if (dp2 == NULL) {
+        return dp1;
+    }
+
+    size_t dp1_size = 0;
+    EFI_DEVICE_PATH_PROTOCOL *node1 = dp1;
+    while (!(node1->Type == END_DEVICE_PATH_TYPE && node1->SubType == END_ENTIRE_DEVICE_PATH_SUBTYPE)) {
+        size_t node_len = node1->Length[0] | (node1->Length[1] << 8);
+        dp1_size += node_len;
+        node1 = (EFI_DEVICE_PATH_PROTOCOL *)((uint8_t *)node1 + node_len);
+    }
+
+    size_t dp2_size = 0;
+    EFI_DEVICE_PATH_PROTOCOL *node2 = dp2;
+    while (!(node2->Type == END_DEVICE_PATH_TYPE && node2->SubType == END_ENTIRE_DEVICE_PATH_SUBTYPE)) {
+        size_t node_len = node2->Length[0] | (node2->Length[1] << 8);
+        dp2_size += node_len;
+        node2 = (EFI_DEVICE_PATH_PROTOCOL *)((uint8_t *)node2 + node_len);
+    }
+    size_t end_node_len = node2->Length[0] | (node2->Length[1] << 8);
+    dp2_size += end_node_len;
+
+    EFI_DEVICE_PATH_PROTOCOL *new_dp = NULL;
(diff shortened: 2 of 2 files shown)


valiukasd:
Hello @iczelia and maintainers.

### **My findings and testing:**

1. **Transparent Compression (`$`)**
Does it work? No. 
Why?
 When you use Limine's compression (like `boot://$efi/boot/bootmgfw.efi.gz`), Limine reads the compressed file from disk and decompresses it into the RAM buffer (which we disable if flag is passed). So, because `NATIVE_LOAD` uses the firmware's native `gBS->LoadImage(..., absolute_path, ...)` without passing a memory buffer, the UEFI firmware bypasses Limine's decompressed RAM buffer entirely. It goes directly to the disk, reads the raw `.gz` bytes, fails to recognize it as a valid PE/COFF executable, and will throw an error (like `EFI_UNSUPPORTED`). **Is it fixable?** Impossible. The entire premise of `NATIVE_LOAD` is to let the motherboard's native drivers load the file from disk so it accurately measures the file path into PCR 4. The firmware has no concept of Limine's software decompression. If a user wants to use `NATIVE_LOAD`, the file must be uncompressed on disk.
2. **Hashing (`#`)**
Does it work? Yes, it will boot, but it introduces a security flaw. 
Why?
When you specify a hash (like a `boot:///efi/boot/bootmgfw.efi#<hash>`), `uri_open()` reads the file into RAM and verifies the hash successfully. Then, `NATIVE_LOAD` calls `gBS->LoadImage()`, which reads the file from the disk a second time. **Is it fixable?** Hard/Impossible without breaking the feature. This creates a Time-Of-Check to Time-Of-Use (**TOCTOU**) vulnerability. Limine verifies the hash of the first read, but UEFI executes the second read. While practically this works for average users, from a strict security standpoint, it doesn't guarantee the exact bytes executed were the ones hashed. Furthermore, it results in the file being read from disk twice (once by Limine, once by UEFI).

What is the best course of action in this case? I suggest we should error out if we try to use compression with `NATIVE_LOAD` and just silently continue if using hashing. But we must add in the documentation that compression does not work, as well as by using hashing with this feature you are vulnerable to this TOCTOU attack.

iczelia:
@valiukasd,

We do not accept undisclosed AI-assisted contributions.

valiukasd:
@iczelia,

Hello. Why is my contribution AI assisted? The code is mine, and I did the testing. I only let AI rewrite my comment to make it easier to read, because I am not native English speaker. I can update my comment to say that it was generated by AI, if that is what is what solves this, but what is AI-assisted? What to say or disclose for this not to happen?

PR #589, opened 2026-06-14
Add support for Physical KASLR

NOTMASTER08:
The limine protocol already doesn't give any guarantees about physical load address, all that's left is to just randomize the physical load address as well if kaslr is enabled.


NOTMASTER08:
Oops, forgot to set the target branch to trunk

PR #590, opened 2026-06-14
Add support for Physical KASLR

NOTMASTER08:
The limine protocol already doesn't give any guarantees about physical load address, all that's left is to just randomize the physical load address as well if kaslr is enabled.


common/mm/pmm.s2.c
```diff
@@ -612,6 +613,109 @@ void *pmm_realloc(void *old_ptr, uint64_t old_size, uint64_t new_size) {
...
+        memmap[i].slot_count = (last_slot - first_slot) / alignment + 1;
+        total_slots += memmap[i].slot_count;
+    }
+
+    if (!total_slots) {
+        return false;
+    }
+    // Allocator potentially calling allocator is risky, but i'll just say it'll
+    // be *fine* for now since this is only used for PKASLR
+    uint64_t slot = (total_slots * (uint64_t)(rand32())) >> 32;
```
iczelia: why did you use this symbol? it's a S3 export and you're modifying S2 code.

NOTMASTER08: Seemed to be the most appropriate option, don't know what to replace with(don't want to roll my own crypto).

iczelia: rand32() is actually not (cryptographically) safe in the slightest; it's Mersenne Twister. hmm...

iczelia: I recommend pulling your fork and using newly exposed HW randomness instead.

NOTMASTER08: Did that, still has the s3/s2 problem, and i don't know how to fix that besides copying the allocation function into pmm.c, or rolling my own PRNG. Neither of which i want to do.

iczelia: Perhaps renaming rand.c to rand.s2.c would help fix this? the MT19937 should go anyway and be replaced by some LCG.

NOTMASTER08: Seems to build now, at least. Not sure if that's the best solution but if it works it works ig.

CherrieTheShifter:
Yes This

common/mm/pmm.s2.c
CherrieTheShifter: Yes

PR #593, opened 2026-06-17
lib/bgrt: natively restore OEM firmware logo upon OS handoff

shelterx:
When Limine restores the terminal before handing off to the OS, the screen is unconditionally wiped via fb_init. By clearing the 'displayed' status bit in the ACPI BGRT table, the OS is forced to natively redraw the firmware logo itself instead of rendering a "blank" OS boot screen.

As previously noted in term_fallback(), this was a known tradeoff: // XXX: Ideally we clear the screen, but that gets rid of the BGRT boot logo

To prevent off-center rendering when Limine negotiates a GOP resolution that differs from the firmware's, the active framebuffer dimensions are used to recalculate and update the BMP image offsets in the BGRT table.

The BMP headers are parsed from the physical image_address using memcpy to avoid strict aliasing violations. It correctly supports both modern BITMAPINFOHEADER and legacy BITMAPCOREHEADER formats, and handles top-down bitmaps via signed height interpretation.

The ACPI table checksum is fully recomputed after all fields are modified, maintaining spec compliance.

This behaviour is entirely gated behind the opt-in firmware_logo: yes configuration flag and is only active on UEFI targets.

Tested on x86_64 UEFI bare metal.

Assisted-by: Antigravity:gemini-3.1-pro

### DIFF
--- a/common/lib/acpi.h
+++ b/common/lib/acpi.h
@@ -199,6 +199,16 @@ struct madt_core_pic {
 #define MADT_CORE_PIC_ENABLED        ((uint32_t)1 << 0)
 #define MADT_CORE_PIC_ONLINE_CAPABLE ((uint32_t)1 << 1)
 
+struct acpi_bgrt {
+    struct sdt header;
+    uint16_t version;
+    uint8_t status;
+    uint8_t image_type;
+    uint64_t image_address;
+    uint32_t image_offset_x;
+    uint32_t image_offset_y;
+} __attribute__((packed));
+
 uint8_t acpi_checksum(void *ptr, size_t size);
 void   *acpi_get_rsdp(void);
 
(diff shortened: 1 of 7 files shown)


shelterx:
Maybe there are other ways to do this but it's an attempt at least.
Unfortunately I don't own any other hardware that supports Limine so I was limited to x86_64 in my testing.

iczelia (changes requested):
Hi,

This looks good but it needs some amendments. Firstly your merge request does not adhere to the code style/contributing guidelines. Please read them thoroughly and make the necessary changes. Secondly, `bgrt->version` is read before `bgrt->header.length < ...` is validated. Thirdly, `(uint8_t *)(uintptr_t)bgrt->image_address` might truncate on 32-bit. I am not sure, I would have to check. For a more overreaching issue, we tend to read config variables in `menu.c`.

shelterx:
Refactored, hopefully it's better now.

iczelia (approved):
All of the issues that I flagged before seem resolved. I will do some final testing and then merge the PR. Thanks.

PR #594, opened 2026-06-18
protos/chainload: Add native_load option for efi protocol

### LINKED ISSUE #576: Support for non-rebooting UEFI handover to prevent PCR alterations

valiukasd:
Fixes and closes #576 

Implements a small code change to load the image via native drivers, if a flag `native_load` is provided. Documentation is updated, no breaking changes for existing users.

Testing status:

- Project builds
- protocol: efi works without the flag `native_load`
- protocol: efi works with the flag `native_load`

Caveats: 

Transparent Compression ($): This does **NOT** work because Limine reads the compressed file from disk and decompresses it into the RAM buffer. Since we skip the RAM buffer with the flag, the UEFI firmware will fail to identify the file a valid PE/COFF executable.

Hashing (#): With the "native_load" flag, the UEFI firmware loads the file again after Limine verifies the hash in RAM. This creates a timing attack, where the file could theoritically be swapped between Limine verifying it, and the UEFI firmware loading the file again. Users of hashing should be aware of this.

Keeping these things in mind, this flag is for Windows dual-boot users where the EFI is rarely hashed or compressed. 

I have update the documentation to add the caveats and also disabled those features if the `native_load` flag is set to `true`. 

> Part of this PR description was generated by AI, subsequent comments from me may be also partly generated by AI

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -287,6 +287,10 @@ Editor control options:
   * `resolution` - The resolution to be used. This setting takes the form of
     `<width>x<height>x<bpp>`. If the resolution is not available, Limine will
     pick another one automatically. Omitting `<bpp>` will default to 32.
+  Here is a shorter, punchier version that cuts the fluff and reads like natural technical documentation:
+  * `native_load` - If set to `yes`, chainloads the EFI application natively without an intermediate memory buffer, ensuring TPM PCR 4 measurements reflect the exact on-disk binary. This is meant for Windows dual-boot users, where Bitlocker trips and asks for a recovery key every on every boot. This disables some features.
+    * **Transparent Compression ($)** is disabled because the RAM buffer used for decompression is bypassed.
+    * **Hashing (#)** is disabled due to a potential Time-of-Check to Time-of-Use (TOCTOU) race condition, where the binary could be swapped between verification and execution.
 
 * EFI Boot Entry protocol:
   * `entry` - The name of the EFI boot entry to reboot into.
--- a/common/protos/chainload.c
+++ b/common/protos/chainload.c
@@ -204,6 +204,44 @@ load:
 
 #elif defined (UEFI)
 
+static EFI_DEVICE_PATH_PROTOCOL *append_device_path(EFI_DEVICE_PATH_PROTOCOL *dp1, EFI_DEVICE_PATH_PROTOCOL *dp2) {
+    if (dp1 == NULL) {
+        return dp2;
+    }
+    if (dp2 == NULL) {
+        return dp1;
+    }
+
+    size_t dp1_size = 0;
+    EFI_DEVICE_PATH_PROTOCOL *node1 = dp1;
+    while (!(node1->Type == END_DEVICE_PATH_TYPE && node1->SubType == END_ENTIRE_DEVICE_PATH_SUBTYPE)) {
+        size_t node_len = node1->Length[0] | (node1->Length[1] << 8);
+        dp1_size += node_len;
+        node1 = (EFI_DEVICE_PATH_PROTOCOL *)((uint8_t *)node1 + node_len);
+    }
+
+    size_t dp2_size = 0;
+    EFI_DEVICE_PATH_PROTOCOL *node2 = dp2;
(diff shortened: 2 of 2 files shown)


iczelia:
@valiukasd,

Judging by the CONFIG.md part of the patch:

```
  Here is a shorter, punchier version that cuts the fluff and reads like natural technical documentation:
  * `native_load` - If set to `yes`, chainloads the EFI application natively without an intermediate memory buffer, ensuring TPM PCR 4 measurements reflect the exact on-disk binary. This is meant for Windows dual-boot users, where Bitlocker trips and asks for a recovery key every on every boot. This disables some features.
    * **Transparent Compression ($)** is disabled because the RAM buffer used for decompression is bypassed.
    * **Hashing (#)** is disabled due to a potential Time-of-Check to Time-of-Use (TOCTOU) race condition, where the binary could be swapped between verification and execution.
```

The patch was also AI-assisted without disclosure in the commit message as required by `CONTRIBUTING.md`, or any type of well-meaning disclosure elsewhere. The disclosure is not optional and this omission shows exceptional lack of care.

Please refrain from submitting pull requests to this repository (in general) in the future.

valiukasd:
@iczelia ,

I am sorry, I have copy pasted the part of the AI comment to the CONFIG.md file. I thought that my footnote in the PR was enough. As I understand, I needed to add Assisted-By to the git commit?

I know I wasted your time with the incorrect PRs, but I have worked hard to create the feature and get familiar with the codebase. I really don't want to keep my own fork, and you are not obliged to listen to me, but can I please submit it one more time? This is one of the first times I contribute to open-source projects, and I acknowleadge that I was in haste to submit the feature, so I overlooked this disclosure.

If you want to still disallow me from opening PRs, anyone reading this is free to take my code and submit it in their own PR.

@iczelia You can do it also.

Have a nice day

PR #599, opened 2026-06-22
protos/linux: Align concatenated initrd modules to 4 bytes

r-vdp:
The Linux initramfs unpacker only accepts a raw newc cpio header when the
current offset in the ramdisk buffer is 4-byte aligned.

Limine currently concatenates `module_path` files byte for byte, so an
uncompressed cpio appended after a compressed initrd whose size is not a
multiple of 4 lands at an unaligned offset and the kernel reports:

```
Initramfs unpacking failed: invalid magic at start of compressed archive
```

This pads each module to a 4-byte boundary and zeroes the gap, matching
GRUB and systemd-boot's EFI stub. Applies to both `linux_x86.c` and
`linux_risc.c`.

Tested on x86_64 UEFI with a zstd-compressed main initrd followed by an
uncompressed newc cpio.

### DIFF
--- a/common/protos/linux_risc.c
+++ b/common/protos/linux_risc.c
@@ -139,7 +139,12 @@ static void load_module(struct boot_param *p, char *config) {
             panic(true, "linux: failed to open module `%s`. Is the path correct?", module_path);
         }
 
-        total_size = CHECKED_ADD(total_size, module_file->size,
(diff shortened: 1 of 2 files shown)


iczelia (approved):
LGTM

PR #604, opened 2026-07-03
lib/getchar: Add configurable keyboard layout support

Ma4lstrom:
Adds support for configurable keyboard layouts via the `keyboard_layout`
configuration option.

This PR implements the Dvorak keyboard layout.

### DIFF
--- a/common/lib/getchar.c
+++ b/common/lib/getchar.c
@@ -5,6 +5,7 @@
 #include <lib/misc.h>
 #include <lib/term.h>
 #include <lib/print.h>
+#include <lib/config.h>
 #if defined (BIOS)
 #  include <lib/real.h>
 #elif defined (UEFI)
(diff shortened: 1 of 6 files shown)


common/lib/getchar.c
```diff
@@ -5,6 +5,7 @@
 #include <lib/misc.h>
 #include <lib/term.h>
 #include <lib/print.h>
+#include <lib/config.h>
```
iczelia: configuration options are typically parsed in menu.c, this include was left out of lib/getchar.c for a specific purpose.

common/lib/getchar.c
```diff
@@ -13,6 +14,36 @@
 #include <drivers/serial.h>
 #include <sys/cpu.h>
 
+
+static int dvorak_enabled = -1;
```
iczelia: Perhaps I suggest turning this into a value of an enum to support more layouts in the future (e.g. AZERTY).

common/lib/getchar.c
```diff
@@ -13,6 +14,36 @@
...
+    ['Q']='\"', ['W']='<', ['E']='>', ['R']='P', ['T']='Y',
+    ['Y']='F', ['U']='G', ['I']='C', ['O']='R', ['P']='L',
+    ['{']='?', ['}']='+', ['|']='|', ['A']='A', ['S']='O',
+    ['D']='E', ['F']='U', ['G']='I', ['H']='D', ['J']='H',
+    ['K']='T', ['L']='N', [':']='S', ['"']='_', ['Z']=':', ['X']='Q',
+    ['C']='J', ['V']='K', ['B']='X', ['N']='B', ['M']='M',
+    ['<']='W', ['>']='V', ['?']='Z',
+};
+
+static bool keyboard_is_dvorak(void) {
```
iczelia: Please move this to menu.c.

test/limine.conf
```diff
@@ -9,6 +9,7 @@ verbose: yes
 wallpaper: ${WALLPAPER_PATH}
 wallpaper_style: centered
 backdrop: 008080
+keyboard_layout: dvorak
```
iczelia: Please remove this line from the test configuration.

common/lib/getchar.c
```diff
@@ -13,6 +14,36 @@
...
+    if(dvorak_enabled == -1) {
+        char *layout = config_get_value(NULL, 0, "keyboard_layout");
```
iczelia: The option is not documented in the respective markdown file.

iczelia (changes requested):
ATM, revisions are necessary (code, documentation and logic-wise). The first two have been addressed in the comments, the logic issue is more overreaching.

Nominally, your patch works on the wrong layer in the UEFI case: `ascii` comes from getchar.c:375, i.e. kd.Key.UnicodeChar; firmware-resolved. The table maps QWERTY -> Dvorak, but this is correct only upon the assumption that the firmware itself considers the keyboard to be US-QWERTY. If this assumption (which is not necessarily true) does not hold, then the behaviour is unintended; e.g. when the user's firmware+keyboard combo already resolves Dvorak keys.

So in principle there is no way to resolve this issue, so it should be documented.

Ma4lstrom:
Hi iczelia,
Thanks for catching the UEFI firmware assumption. I've documented it as a known limitation in CONFIG.md, along with the new keyboard_layout option itself.
For sharing that state with getchar.c, I went with a global (current_keyboard_layout, exposed via menu.h). If you'd prefer a different approach there, let me know and I'll happily change it.
I’ve also made adjustments to your other comments as well.

iczelia (approved):
LGTM

PR #607, opened 2026-07-09
Add Dvorak number row and AZERTY keyboard layout support

mrunix00:
- Limine used to only map letters from QWERTY to DVORAK, but not the symbols and the number row, the commit 98b7e67ab3577fa292bca0073a1e82ab8cc46634 fixes that

| QWERTY | DVORAK |
| --- | --- |
| image | image |

- Added support for AZERTY layout in the commit 55dc11c3cf0a3cca070d934bf5896ddf1e8e979c
- Extended ASCII characters such as `éèçà` were getting skipped by `isprint`, now they're included in this commit: 98b7e67ab3577fa292bca0073a1e82ab8cc46634, I should note that [flanterm](https://github.com/Mintsuki/Flanterm) still doesn't support printing extended ASCII characters so they get rendered as squares


iczelia (approved):
LGTM, thanks!

PR #615, opened 2026-07-23
Allow Limine base revision and requests end marker to be at very end of data section

benjidial:


benjidial:
Oops, I targeted the wrong branch. I'll close this request and open a new one with the correct branch. (Is there an easier way to apply the same change to trunk instead?)

PR #619, opened 2026-07-31
common/mm/pmm.s2: Handle reserved regions inside usable memory

aba2222:
Split off usable chunks around reserved regions.

This removes a case where valid memory maps could trigger a panic during in pmm_sanitise_entries.


CherrieTheShifter:
Yes This

Issue #611, opened 2026-07-19
FR: Allow for multi-line `cmdline` params with backslashes.


VenomousSteam81:
It would be really nice if you could have a fairly readable config file. From this mess:

```
cmdline: root=UUID=bcb5ce41-9013-49e6-940d-8d5bc8c586fd rw rootfstype=f2fs loglevel=11 splash zswap.enabled=0
```

to:

```
cmdline: root=UUID=bcb5ce41-9013-49e6-940d-8d5bc8c586fd rw rootfstype=f2fs loglevel=11 splash \
    zswap.enabled=0
```

Mobeen0119:
I'm working on this. The change goes in init_config() in common/lib/config.c, right after the trailing newline is added and before the existing whitespace cleanup. When a \ is immediately followed by \n, both characters get stripped, joining the next line to the current one. I'll submit a PR once tested.

CherrieTheShifter:
Yes This

PR #625, opened 2026-08-06
config: support backslash-newline line continuation (#611)

Mobeen0119:
**Add backslash-newline line continuation to config parser.**

O(n) two-pointer approach.
Handles Unix (\n) and Windows (\r\n) line endings.

**Example:**
  CMDLINE: root=/dev/sda1 rw quiet \
      zswap.enabled=0

Becomes one continuous line.

Tested with 20 unit tests + 100K fuzz + 50K comparisons vs old code.

**_Fixes #611_**

**Test Files :** 
[test_compare.c](https://github.com/user-attachments/files/30767339/test_compare.c)
[test_full.c](https://github.com/user-attachments/files/30767340/test_full.c)

### DIFF
--- a/common/lib/config.c
+++ b/common/lib/config.c
@@ -409,34 +409,54 @@ int init_config(size_t config_size) {
 
     // add trailing newline if not present
     config_addr[config_size - 2] = '\n';
-
+    
     size_t config_alloc_size = config_size;
 
-    // remove windows carriage returns and spaces at the start and end of lines, if any
-    for (size_t i = 0; i < config_size; i++) {
-        size_t skip = 0;
-        if (config_addr[i] == ' ' || config_addr[i] == '\t') {
-            while (i + skip < config_size && (config_addr[i + skip] == ' ' || config_addr[i + skip] == '\t')) {
-                skip++;
+    // handle backslash-newline line continuation
+    size_t write = 0;
+    for (size_t read = 0; read < config_size; read++) {
+        if (config_addr[read] == '\\' && read + 1 < config_size) {
+            if (config_addr[read + 1] == '\n') {
+                read++;
+                continue;
             }
-            if (i + skip < config_size && config_addr[i + skip] == '\n') {
-                goto skip_loop;
+            if (config_addr[read + 1] == '\r' && read + 2 < config_size && config_addr[read + 2] == '\n') {
+                read += 2;
+                continue;
             }
-            skip = 0;
         }
-        while (i + skip < config_size
-            && ((config_addr[i + skip] == '\r')
-                || ((!i || config_addr[i - 1] == '\n') && (config_addr[i + skip] == ' ' || config_addr[i + skip] == '\t')))
-        ) {
-            skip++;
+        config_addr[write++] = config_addr[read];
+    }
+
+    config_size = write;
+
+    // Remove carriage returns and leading/trailing whitespace from lines
+    {
+        size_t trim_write = 0;
+        bool at_line_start = true;
+        for (size_t read = 0; read < config_size; read++) {
+            char c = config_addr[read];
+
+            if (c == '\r') {
+                continue;
+            }
+
+            if (at_line_start && (c == ' ' || c == '\t')) {
+                continue;
+            }
+
+            if (c == '\n') {
+                while (trim_write > 0 && (config_addr[trim_write - 1] == ' ' || config_addr[trim_write - 1] == '\t')) {
+                    trim_write--;
+                }
+                at_line_start = true;
+            } else {
+                at_line_start = false;
+            }
+
+            config_addr[trim_write++] = c;
         }
-skip_loop:
-        if (skip) {
-            for (size_t j = i; j < config_size - skip; j++)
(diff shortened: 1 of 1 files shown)


programmerlexi:
This is the third time you have created this PR. Can you please just amend/rebase new changes and force push them instead of recreating the PR everytime?

CherrieTheShifter:
Yes This Please 🙏

Zesko:
I would suggest improving performance when `limine.conf` contains many lines.

It should skip lines that don't contain kernel command lines instead of checking every single line. 
Otherwise, checking all lines can unnecessarily increase boot time on large configurations.

Mobeen0119:
This pipeline already runs multiple full-buffer passes before mine (hash check, continuation-collapse, whitespace trim, macro load, macro expand), so mine isn't adding a new category of cost and skipping non-cmdline lines isn't possible at this point since nothing's tokenized into lines yet. Worth noticing: the whitespace-trim loop right after mine is actually O(n²), not O(n) ... it shifts the entire remaining buffer on every skip event, which is the bigger perf risk on large configs, and it's pre-existing.

hotline1337 (commented):
> This pipeline already runs multiple full-buffer passes before mine (hash check, continuation-collapse, whitespace trim, macro load, macro expand), so mine isn't adding a new category of cost and skipping non-cmdline lines isn't possible at this point since nothing's tokenized into lines yet. Worth noticing: the whitespace-trim loop right after mine is actually O(n²), not O(n) ... it shifts the entire remaining buffer on every skip event, which is the bigger perf risk on large configs, and it's pre-existing.

Since you already identified the real performance concern - it might be worth fixing it in the same PR. Right now it shifts the entire remaining buffer on every skip event:

```c
if (skip) {
    for (size_t j = i; j < config_size - skip; j++)
        config_addr[j] = config_addr[j + skip];
    config_size -= skip;
    i--;
}
```

With `k` skip events this is O(n*k), worst case O(n^2) (e.g. a config full of trailing spaces).

Here's a drop-in O(n) replacement using the same read/write pointer pattern as your continuation pass:

```c
    // Remove carriage returns and leading/trailing whitespace from lines
    {
        size_t write = 0;
        bool at_line_start = true;
        for (size_t read = 0; read < config_size; read++) {
            char c = config_addr[read];

            if (c == '\r') {
                continue;
            }

            if (at_line_start && (c == ' ' || c == '\t')) {
                continue;
            }

            if (c == '\n') {
                while (write > 0 && (config_addr[write - 1] == ' ' || config_addr[write - 1] == '\t')) {
                    write--;
                }
                at_line_start = true;
            } else {
                at_line_start = false;
            }

            config_addr[write++] = c;
        }
        config_size = write;
    }
```

Same semantics: strips `\r`, strips leading whitespace after `\n`, strips trailing whitespace before `\n`. The trailing-whitespace backtrack is amortised O(1) per character (each byte is "backed over" at most once). Single pass, no buffer shifts.

Mobeen0119:
old code kept trailing spaces before \r\n but stripped them before \n  which is inconsistent. this code strips them the same way either time, which is correct. Tested against 200k fuzz inputs + 30 edge cases to confirm this is the only real difference from the old behavior. wanted to call it out since it's a behavior change, not just a speed fix, before merging.

common/lib/config.c
```diff
@@ -410,33 +410,53 @@ int init_config(size_t config_size) {
...
+                read += 2;
+                continue;
+            }
+        }
+        config_addr[write++] = config_addr[read];
+    }
+
+    config_size = write;
+
     size_t config_alloc_size = config_size;
```
hotline1337: This is now computed after the continuation-collapsing loop has already reduced `config_size`. In the current code, `config_alloc_size` is captured right after L#441 (setting the trailing newline), before any size-reducing pass.

hotline1337 (changes requested):
Move ``config_alloc_size = config_size;`` to before the continuation pass (i.e., right after L#411, before L#414), otherwise `pmm_free(config_addr, config_alloc_size)` will be called with the wrong size whenever the config uses line continuation at all, which is the entire point of this feature.

common/lib/config.c
```diff
@@ -409,34 +409,56 @@ int init_config(size_t config_size) {
...
-        }
-skip_loop:
-        if (skip) {
-            for (size_t j = i; j < config_size - skip; j++)
-                config_addr[j] = config_addr[j + skip];
-            config_size -= skip;
-            i--; // re-examine character shifted into position i
+        config_addr[write++] = config_addr[read];
+    }
+
```
hotline1337, line 430: Extra blank lines here don't add readability. Removing them tightens up the block between the loop and the next processing step.

hotline1337 (changes requested):
I'd suggest formatting this section in this way:
```
    }

    config_size = write;

    // Remove carriage returns and leading/trailing whitespace from lines
```

**eelgtm!**

hotline1337 (approved):
Good Job! Now it's up to the maintainers if this gets approved to be merged. Thank you for fixing and improving the code!

PR #628, opened 2026-08-12
mm/pmm: Bound UEFI page allocation requests

tomalok:
## Summary

Bound the size of UEFI `AllocatePages()` requests made while claiming usable
physical memory.

Requests initially use a maximum chunk size of 1 GiB. If firmware rejects a
request, the cap is reduced adaptively, down to a minimum of 64 MiB. A failed
chunk is then retried in 1 MiB pieces so that usable memory can still be
recovered on small or fragmented systems.

Adjacent pieces which cannot be claimed through UEFI are coalesced into
`MEMMAP_EFI_RECLAIMABLE` ranges.

## Motivation

Some UEFI implementations fail or stall when an `AllocatePages()` request
spans a large fraction of physical memory.

The existing fallback retries a rejected range one 4 KiB page at a time. On a
large-memory system, that can require tens of millions of firmware calls and
make the machine appear to hang during boot.

Bounding the normal request size avoids problematic very large allocations,
while the adaptive reduction accommodates more restrictive firmware. The
1 MiB recovery path retains the existing ability to recover fragmented memory
without falling back to individual-page requests.

This is separate from #619: that change handles nested reserved regions while
sanitising the firmware memory map, whereas this change bounds the subsequent
UEFI `AllocatePages()` requests used to claim sanitised usable regions.

## Testing

- Built all five UEFI targets with warnings treated as errors:
  - IA-32
  - x86_64
  - AArch64
  - RISC-V 64
  - LoongArch 64
- Exercised the allocation policy and failure paths with a test harness.
- Booted the Limine protocol test under UEFI with 512 MiB:
  - x86_64 QEMU
  - AArch64 QEMU
- Completed Alpine Linux packaged-image boot regressions using
  `limine-12.5.2-r4`.
- Booted an Alpine AArch64 image on AWS `t4g.nano` with 512 MiB.
- Booted the same packaged Limine build on AWS `m6g.metal` with 256 GiB.
  Linux detected the complete 256 GiB physical-memory total, enabled EFI
  runtime services, and brought all 64 CPUs online.

### DIFF
--- a/common/mm/pmm.s2.c
+++ b/common/mm/pmm.s2.c
@@ -312,6 +312,82 @@ static struct memmap_entry *recl;
 
 extern symbol __slide, __image_base, __image_end;
 
+// Some UEFI implementations cannot handle allocations spanning a large
+// fraction of physical memory. Reduce the request cap no lower than 64MiB,
+// and use 1MiB requests only to recover a failed chunk.
+#define UEFI_ALLOC_MAX_PAGES      ((UINTN)(0x40000000 / PAGE_SIZE))
+#define UEFI_ALLOC_MIN_PAGES      ((UINTN)(0x04000000 / PAGE_SIZE))
+#define UEFI_ALLOC_FALLBACK_PAGES ((UINTN)(0x00100000 / PAGE_SIZE))
+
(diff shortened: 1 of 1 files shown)


tomalok:
@Mintsuki

PR #630, opened 2026-08-13
lib/getchar: Bound BIOS waits with a TSC deadline

tomalok:
## Description

BIOS menu delays currently depend on IRQ0 advancing the PIT tick counter. If those interrupts are not delivered, `_pit_sleep_and_quit_on_keypress()` can wait indefinitely.

Retain the existing PIT-based timing and input handling, but also calculate a deadline using the already calibrated TSC. The real-mode wait loop exits when either the requested PIT tick count or the TSC deadline is reached.

A zero deadline represents an unavailable TSC frequency and preserves the existing PIT-only behavior.

This does not change UEFI builds or BIOS serial-port discovery.

## Testing

- Built the BIOS code with warnings treated as errors.
- Verified an automatic menu timeout under QEMU without serial input.
- Verified an automatic menu timeout with COM1 enabled.
- Verified that keyboard input still interrupts the timeout.
- Verified with an instrumented QEMU build that the timeout completes using the TSC deadline when the IRQ0 tick counter is prevented from advancing.

### DIFF
--- a/common/lib/getchar.c
+++ b/common/lib/getchar.c
@@ -175,7 +175,8 @@ int getchar_internal(uint8_t scancode, uint8_t ascii, uint32_t shift_state) {
 }
 
 #if defined (BIOS)
-int _pit_sleep_and_quit_on_keypress(uint32_t ticks, uint32_t aux_poll);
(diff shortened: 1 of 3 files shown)


tomalok:
@Mintsuki

PR #631, opened 2026-08-13
drivers/serial: Discover the BIOS console UART

tomalok:
# drivers/serial: Discover the BIOS console UART

> [!IMPORTANT]
> Depends on #630. This branch is currently stacked on that PR, and only the second commit is new here. This PR will be rebased onto `trunk` after #630 merges.

## Description

The BIOS serial driver currently assumes COM1 is available at I/O port `0x3f8`. If the configured device is absent or unresponsive, waiting for its transmitter can block boot indefinitely.

Discover the console UART from the first serial-port entry in the BIOS Data Area, falling back to the ACPI Serial Port Console Redirection table when the BDA entry is empty.

The SPCR path supports byte-wide, one-byte-stride 16550-compatible interfaces using either system I/O or memory-mapped I/O. Supported interface types are:

- `0`: fully 16550-compatible
- `1`: 16450-compatible with 16550 FIFO-control support
- `0x12`: revision 2 or later 16550-compatible interface described by the Generic Address Structure

Unsupported or invalid SPCR descriptions disable serial output instead of accessing an assumed COM1 device.

Serial transmission is also bounded by the TSC deadline helpers introduced in #630. If a discovered UART does not become ready, serial output is disabled and boot continues.

A quiet BIOS ACPI-table lookup is included because ordinary ACPI lookup diagnostics may themselves use serial output, which would recursively re-enter UART initialization.

## Testing

- Built all Limine targets with warnings treated as errors.
- Verified ordinary BDA/COM1 discovery and serial output under QEMU.
- Verified boot with serial disabled in the configuration.
- Verified `SERIAL=yes` continues booting when no UART is exposed.
- Forced SPCR discovery under QEMU and verified interfaces `0`, `1`, and `0x12`.
- Verified SPCR serial menu output, automatic timeout, and test-kernel boot.
- Backported both stacked changes to Limine 12.5.2 and built them successfully.
- Verified Alpine `limine-12.5.2-r5` on AWS `t3a.nano`.
- Verified Alpine `limine-12.5.2-r5` boots successfully on AWS `m5.metal`.

### DIFF
--- a/common/drivers/serial.c
+++ b/common/drivers/serial.c
@@ -2,30 +2,110 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <stddef.h>
+#include <lib/acpi.h>
 #include <lib/misc.h>
 #include <drivers/serial.h>
 #include <sys/cpu.h>
 
 static bool serial_initialised = false;
+static bool serial_mmio;
+static uintptr_t serial_base;
 uint32_t serial_baudrate;
+
+struct acpi_gas {
+    uint8_t address_space;
+    uint8_t bit_width;
+    uint8_t bit_offset;
+    uint8_t access_size;
+    uint64_t address;
+} __attribute__((packed));
+
+struct acpi_spcr {
+    struct sdt header;
+    uint8_t interface_type;
+    uint8_t reserved[3];
+    struct acpi_gas base_address;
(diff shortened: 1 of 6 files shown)


tomalok:
@Mintsuki

PR #648, opened 2026-09-06
limine: Allow EL2 without VHE to boot on EL1 under base revision 6

YiraSan:
## Rationale

Currently, Limine triggers a hard panic on base revision 6 when booting on AArch64 systems running at EL2 without VHE (Virtualization Host Extensions). While investigating this restriction, it appeared unclear why non-VHE EL2 environments were outright excluded, as they affect a massive portion of the ARM ecosystem.

Architecturally, Virtualization Host Extensions (`FEAT_VHE`) were only introduced starting with ARMv8.1-A. A vast majority of foundational ARMv8-A hardware implements ARMv8.0 and therefore fundamentally lacks hardware VHE support. For instance, the Raspberry Pi 4 is an ARMv8.0 SoC. When booted via standard UEFI firmware layers, control is handed over to the bootloader at EL2 without VHE.

While developers could theoretically fall back to older revisions, doing so means missing out on all the critical bug fixes, enhanced CPU state guarantees during kernel handoff, etc... It feels a bit unfortunate to leave out ubiquitous hardware when a graceful EL2-to-EL1 transition path is feasible.

## Proposed Changes

This pull request introduces a clean, reliable EL2-to-EL1 transition path for non-VHE environments under base revision 6, including support for MpRequest.

This modification is entirely non-breaking, while it will require to slightly modify the protocol specification. Any platform or environment that successfully booted under base revision 6 prior to this change will continue to do so identically.

I have tested this PR on my own Raspberry Pi 4B.

### DIFF
--- a/common/lib/spinup.asm_aarch64
+++ b/common/lib/spinup.asm_aarch64
@@ -5,10 +5,41 @@
 // noreturn void enter_in_el1(uint64_t entry, uint64_t sp, uint64_t sctlr,
 //                            uint64_t mair, uint64_t tcr, uint64_t ttbr0,
 //                            uint64_t ttbr1, uint64_t direct_map_offset)
-// Configure EL1 state and jump to kernel. Must be called at EL1.
+// Configure EL1 state and jump to kernel.
 
 .global enter_in_el1
 enter_in_el1:
+    // Drop EL2 -> EL1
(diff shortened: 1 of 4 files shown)


YiraSan:
Damn I forgot to target trunk. I will fix that.

PR #649, opened 2026-09-06
common: Allow EL2 without VHE to boot on EL1 under base revision 6

YiraSan:
## Rationale

Currently, Limine triggers a hard panic on base revision 6 when booting on AArch64 systems running at EL2 without VHE (Virtualization Host Extensions). While investigating this restriction, it appeared unclear why non-VHE EL2 environments were outright excluded, as they affect a massive portion of the ARM ecosystem.

Architecturally, Virtualization Host Extensions (`FEAT_VHE`) were only introduced starting with ARMv8.1-A. A vast majority of foundational ARMv8-A hardware implements ARMv8.0 and therefore fundamentally lacks hardware VHE support. For instance, the Raspberry Pi 4 is an ARMv8.0 SoC. When booted via standard UEFI firmware layers, control is handed over to the bootloader at EL2 without VHE.

While developers could theoretically fall back to older revisions, doing so means missing out on all the critical bug fixes, enhanced CPU state guarantees during kernel handoff, etc... It feels a bit unfortunate to leave out ubiquitous hardware when a graceful EL2-to-EL1 transition path is feasible.

## Proposed Changes

This pull request introduces a clean, reliable EL2-to-EL1 transition path for non-VHE environments under base revision 6, including support for MpRequest.

This modification is entirely non-breaking, while it will require to slightly modify the protocol specification. Any platform or environment that successfully booted under base revision 6 prior to this change will continue to do so identically.

I have tested this PR on my own Raspberry Pi 4B.

### DIFF
--- a/common/lib/spinup.asm_aarch64
+++ b/common/lib/spinup.asm_aarch64
@@ -5,10 +5,41 @@
 // noreturn void enter_in_el1(uint64_t entry, uint64_t sp, uint64_t sctlr,
 //                            uint64_t mair, uint64_t tcr, uint64_t ttbr0,
 //                            uint64_t ttbr1, uint64_t direct_map_offset)
-// Configure EL1 state and jump to kernel. Must be called at EL1.
+// Configure EL1 state and jump to kernel.
 
 .global enter_in_el1
 enter_in_el1:
+    // Drop EL2 -> EL1
+    mrs x8, currentel
+    cmp x8, #0b1000 // EL2 ?
+    b.ne 0f 
+
+    msr sp_el0, x1
+
+    msr mair_el1, x3
+    msr tcr_el1, x4
+    msr ttbr0_el1, x5
+    msr ttbr1_el1, x6
+    msr sctlr_el1, x2
+    isb
+    dsb sy
+    isb
+
+    tlbi vmalle1
+    dsb nsh
+    isb
+
+    mov x8, #(1 << 31)
+    msr hcr_el2, x8
+
+    mov x8, #0x3c4
+    msr spsr_el2, x8
+
+    msr elr_el2, x0         
+
+    eret
+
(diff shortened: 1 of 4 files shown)


Mintsuki:
Hi, thanks for the PR.

Besides several style concerns with the PR, this method of boot (EL2 to EL1 drop) existed, and was ditched for a reason.

The method would require bootloader trapping guarantees which require setting a mask of bits of what *not* to trap, and these bits change over time as aarch64 and its implementations evolve. Basically rendering aarch64 not backwards compatible (this path in particular).

Please direct your complaints to ARM Holdings plc.

YiraSan:
Hi, thanks for taking the time to review the PR.

I completely understand the general concern regarding the maintenance burden of trap masks and forward compatibility as aarch64 evolves.

I would just like to clarify one architectural detail: this specific non-VHE EL2-to-EL1 fallback path is exclusively limited to ARMv8.0. According to the ARM Architecture Reference Manual, from ARMv8.1 onwards, FEAT_VHE is strictly mandatory whenever EL2 is implemented. Therefore, any newer CPU will inherently take Limine's existing native VHE boot path.

Because ARMv8.0 is a finalized and frozen specification, the trap mask required for this specific non-VHE drop is entirely static. It is immune to future aarch64 extensions, meaning this specific path would never actually require the ongoing "whack-a-mole" updates you rightfully want to avoid.

That being said, I fully respect the project's strict design philosophy. I will happily maintain this patch in my downstream fork if you decide not to re-open this PR.

*Looking back at my own patch, I realize I actually forgot to disable trapping bits... 😅*

Mintsuki:
Thanks for the kind suggestions. I made it work in 685ce5fbf000a7cfecbf6e53cd05e8e2e40f8cc6.

PR #651, opened 2026-09-07
menu: add BOOT_KEYBIND for direct entry booting

viewerofall:
Adds a global `BOOT_KEYBIND: <key> <index>` config option that binds a
single key to boot into the given entry based on its place in a flat list, not expanded meaning directories are just counted as another one so everything under them is just another one.

### Behaviour

- Case-insensitive on both sides (config value and keypress), so `H` and
  `h` are equivalent.
- Respects `keyboard_layout` remapping — the bound key is matched against
  the post-remap character, same as every other menu keystroke.
- Cannot target a sub-menu/directory entry, only a bootable entry. 
- Keys that are hardcoded or already bound such as (`1`-`9`, `e`, `s`, `u`, `b`) will make the keybind be skipped, giving the hardcoded ones priority
- If the same key is bound more than once, the one listed first wins.
- Any malformed line is skipped silently allowing you to have a successful go with your config, just a dysfunctional line. 
- Index follows a flat list, not a tree one you do not have to expand directories in order to boot your entry.

### Testing

- Built and booted under QEMU (BIOS and UEFI) against a minimal limine protocol setup over 3 entries
- Verified on real hardware (UEFI boot).
- Confirmed keybinds fire during the timeout countdown as well as after
  it's aborted (single keypress does both).

### Docs

`CONFIG.md` updated with the new `BOOT_KEYBIND` option under the
Miscellaneous global options list.

---

Assisted-by: Claude:claude-sonnet-5


viewerofall:
@Mintsuki can you see why this is failing, because I cannot get this to replicate anywhere on my system. 
Ive recloned it twice, ive configured it differently. I cant get this thing to do the error to figure out where the hell I am going wrong. `bootsect.asm:165: error: TIMES value -1 is negative` is what is happening in the workflow and I cant figure out what happened to change that. Is it tolower using more storage? 

EDIT: After checking, this seems to be caused be your stage1/hdd commit changing and eating up more bytes that it should causing it to crash with an out of storage error practically

viewerofall:
@Mintsuki when you are done with everything and such, can you review or comment then merge this. Its been a week and I dont want any code editing this making it not work. Thank you.

PR #661, opened 2026-09-13
Limine showing UNVERSIONED when building from a git submodule

cbpudding:
Good afternoon,

While working on a Linux distribution, I ran into an issue where version.sh did not believe that it was operating in a git repository. Since a submodule creates .git as a file, the check fails, and the bootloader shows UNVERSIONED. Switching this to simply check if the file exists, regardless of mode, fixes the issue.

Hope this helps!
~Alex


cbpudding:
PR not targeting trunk and the fork GitHub made doesn't have the trunk branch. Closing this and forking again...

